PR #78, opened 2020-03-25
Feature/prior arithmetic

rhayes777:
All these operations are now possible between priors and floats:
prior = a + b
prior = a - b
prior = a * b
prior = a / b
prior = a // b
prior = a % b
prior = abs(a)
prior = a < b

Any combination is also allowed.


rhayes777:
- The same also applies to annotation and promises.
- A bug where values compared in assertions would be converted to physical values twice over is fixed.
- JSON config now always indents properly.

Issue #82, opened 2020-04-07
Changes to support multiple NLOs


Jammy2211:
I am going to begin implementing more NLOs.

First, I think we need to change the path structure for outputs. Currently, these go something like:

> pipeline_name
> ----phase_name
> ---------phase_tag
> ---------------optimizer-(this-is-the-sym-link)
> ---------------------multinest-chains
> ---------------optimizer_backup
> ---------------------multinest-chain-backups
> ---------------images
> --------------------multinest-best-fit-images,-etc.
> ---------------model.info
> ---------------model.results
> ---------------metadata1
> ---------------data.pickle
> ---------------mask.pickle

For multiple NLOs, we want these to all be in a folder named after the nlo, I suggest something like this:

> pipeline_name
> ----phase_name
> ---------phase_tag
> ---------------multinest
> ---------------------chains-(this-is-the-sym-link)
> ---------------------------multinest-chains
> ----------------------chains_backup
> ---------------------------multinest-chain-backups
> ----------------------images
> -------------------------------multinest-best-fit-images,-etc.
> ----------------------aggregator
> ------------------------------metadata
> ------------------------------data.pickle
> ------------------------------mask.pickle
> ---------------model.info
> ---------------model.results

We should also add the non linear class to the metadata file and allow us to filter for non linear class (or some sort of string) in the aggregator.

Could we also preserve BC such that if an existing run sees the optimizer layout it converts it to the new multinest layout (you can assume only multinest is used for conversion).

Jammy2211:
Configs should be updated to change the non_linear.ini to a folder of configs for each NLO.

Jammy2211:
The sym link is only necessary for Multinest - shall we make it an optional input of the NLO (its a huge pain for HPC file storage and whatnot).

Jammy2211:
I have renamed files such that are output as follows:

multinest.paramnames -> nls.paramnames
multinest_phys_live_points -> nls_phys_live_points

etc.

Jammy2211:
Implemented.

PR #89, opened 2020-04-10
Nlo paths

Jammy2211:
Attempted to sort out the paths for this issue:

https://github.com/rhayes777/PyAutoFit/issues/82

Was unable to:

- Put the pickles in a folder - had issues making the path.
- Make a MulitNestPaths class.


autofit/optimize/non_linear/downhill_simplex.py
```diff
@@ -18,7 +18,7 @@ def __init__(
             fmin=scipy.optimize.fmin
     ):
         if paths is None:
-            paths = Paths()
+            paths = Paths(non_linear_name=self.name)
```
rhayes777, line 21: Let's use the name of the class rather than have a property specifically

autofit/optimize/non_linear/paths.py
```diff
@@ -45,13 +45,19 @@ def wrapper(self, *args, **kwargs):
 
         remove_files = conf.instance.general.get("output", "remove_files", bool)
 
+        # TODO : This is a hack you'll need to fix. I can acces the name by instantiating the NonLinearClass, but I have
+        # TODO : to make sure patths are not created by setting it to an empty string.
+
+        non_linear_name = kwargs["non_linear_class"](paths="").name if "non_linear_class" in kwargs else ""
```
rhayes777, line 51: kwargs.get("non_linear_class", "")

PR #103, opened 2020-05-13
Feature/eden

rhayes777:
- Makes a copy of the project in the same directory suffixed with _eden
- Replaces references that use the project abbreviation with explicit paths in tests
- Wipes the top level \_\_init\_\_

```bash
python edenise.py /path/to/autofit autofit af
```


rhayes777:
All \_\_init\_\_ imports must be relative (e.g. from .module rather than from autofit.module)

Wildcard imports are not allowed

PR #110, opened 2020-07-22
Feature/message passing

rhayes777:
Here's an initial message passing test, refactor and experiment.

I'd like you guys to take a look as a sanity check. There's a specific test I've written which attempts to model multiple gaussians that share a common intensity variable. It seems to work and should illustrate a first step towards the kind of hierarchical modelling we're aiming to achieve.

I have a few ideas in mind for how to improve the API. I largely want to make more components implicit.

James - it might be a good idea to extend my gaussian work and try to fit some more stuff. Although I may change the API and make it more autofitty
Matthew - Could you implement termination for my LaplaceOptimiser ?


test_autofit/unit/message_passing/gaussian/test_multiple.py
```diff
@@ -0,0 +1,209 @@
+from typing import Dict
```
rhayes777, line 1: This is the file in which I attempt a shared-variable model

autofit/message_passing/messages/abstract.py
```diff
@@ -0,0 +1,256 @@
...
+        natural = self.natural_parameters
+        new_params = other * natural
+        log_norm = other * self.log_norm
+        return self.from_natural_parameters(new_params, log_norm=log_norm)
+
+    def __repr__(self):
+        return f"{type(self).__name__}({self}"
+
+    def __str__(self):
+        return ""
```
matthewghgriffiths, line 137: Why did you remove the formatting of the repr? It can be quite useful when debugging in interactive mode? I'm guessing the linter didn't like it?

rhayes777: Good question. I think I actually meant to fix it afterwards but it got in the way of refactoring something else. Will see if I can put it back

PR #119, opened 2020-09-09
Feature/message passing

rhayes777:
Renamed it to graphical in the end.

This is basically what I showed you earlier with some additional documentation. I'm about to go unify variables and priors. Wish me luck.

@matthewghgriffiths did you ever notice that you look a bit like the guy from Veritaseum?


test_autofit/unit/graphical/gaussian/test_declarative.py
```diff
@@ -0,0 +1,246 @@
+import numpy as np
+import pytest
+
+import autofit as af
+import autofit.graphical as ep
+from test_autofit.unit.graphical.gaussian.model import Gaussian, make_data, _likelihood
+
+
+def test_shared_intensity():
```
rhayes777, line 9: @Jammy2211 this the test to use as an example for a notebook.

PR #144, opened 2020-12-10
Ordering of columns does not matter when loading samples

rhayes777:


autofit/non_linear/optimize/abstract_optimize.py
```diff
@@ -1,28 +1,24 @@
...
 
 from autoconf import conf
-
-from autofit.non_linear.abstract_search import NonLinearSearch
 from autofit.non_linear import samples as samp
+from autofit.non_linear.abstract_search import NonLinearSearch
+
 
-class AbstractOptimizer(NonLinearSearch):
+class AbstractOptimizer(NonLinearSearch, ABC):
```
rhayes777, line 8: If a class does not implement all the abstract methods of its parent(s) it should be marked as ABC

autofit/non_linear/samples.py
```diff
@@ -9,38 +10,48 @@
...
+    output = defaultdict(list)
 
     with open(filename, "r+", newline="") as f:
         reader = csv.reader(f)
-        for index, row in enumerate(reader):
-            if index > 0:
-                parameters.append(
-                    [float(param) for param in row[0: model.prior_count]]
+        headers = next(reader)
+        for row in reader:
```
rhayes777, line 19: Each row ought to be a "Sample".

Jammy2211 (approved):

PR #150, opened 2021-01-21
JOSS paper edits

dfm:
This PR includes the updated citation for NumPy.

But I also have a few questions:

- On line 29 of [the manuscript](https://raw.githubusercontent.com/openjournals/joss-papers/joss.02550/joss.02550/10.21105.joss.02550.pdf), it's not clear to me what "linear equations" means. It's certainly not true that existing PPLs only support linear models. Please clarify this in the manuscript!
- Similarly, on line 35, the claim that PyAutoFit is more suitable for "longer term software projects" should be clarified. Existing PPL frameworks are widely used as backends for production software or other long term software projects.

Ref: https://github.com/openjournals/joss-reviews/issues/2550

### DIFF
--- a/paper/paper.bib
+++ b/paper/paper.bib
@@ -1,12 +1,26 @@
-@ARTICLE{numpy,
-  author={S. {van der Walt} and S. C. {Colbert} and G. {Varoquaux}},
-  journal={Computing in Science   Engineering},
-  title={The NumPy Array: A Structure for Efficient Numerical Computation},
-  doi={10.1109/MCSE.2011.37},
-  year={2011},
-  volume={13},
-  number={2},
-  pages={22-30},}
+@Article{         numpy,
+ title         = {Array programming with {NumPy}},
+ author        = {Charles R. Harris and K. Jarrod Millman and St{'{e}}fan J.
+                 van der Walt and Ralf Gommers and Pauli Virtanen and David
+                 Cournapeau and Eric Wieser and Julian Taylor and Sebastian
+                 Berg and Nathaniel J. Smith and Robert Kern and Matti Picus
+                 and Stephan Hoyer and Marten H. van Kerkwijk and Matthew
+                 Brett and Allan Haldane and Jaime Fern{'{a}}ndez del
+                 R{'{\i}}o and Mark Wiebe and Pearu Peterson and Pierre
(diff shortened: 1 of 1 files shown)


Jammy2211:
> On line 29 of the manuscript, it's not clear to me what "linear equations" means. It's certainly not true that existing PPLs only support linear models. Please clarify this in the manuscript!

I remember struggling with writing the right phrase for this sentence. I'm trying to express how for many PPL's the typical use case is that the model is normally something like a linear regression line, or a Gaussian distribution, or something where the model definition is clear. This contrasts the more abstract models we use PyAutoFit to build, which represent things like the mass of galaxies.

If you have any suggestions for a better phrase (or think what I'm trying to get at isn't reasonable) let us know, I'm going to think it over a bit more.

> Similarly, on line 35, the claim that PyAutoFit is more suitable for "longer term software projects" should be clarified. Existing PPL frameworks are widely used as backends for production software or other long term software projects.

It is the "more complete management of the model fitting task" we are trying to use as justification for this. For example, the model-fitting procedure in PyAutoFit can be linked to the visualization routines of the software, such that it outputs visualization for the method during the model-fit. It Also handles something like the result output file structure more comprehensively than other PPLs, as far as I'm aware. To my knowledge this level of interface between PPL and software isn't available in another PPL (albeit I could well be wrong!).  

Nevertheless, the phrase "more complete management of the model fitting task than other PPLs " is probably a bit much and feels uncessary, so I will rephrase this to avoid the comparison to other PPLs.

Jammy2211:
My rephrasings are below, I had trouble getting git to recognise the PR so have pushed to master + develop.

Line 29:

`Each package is specialized to problems of a certain nature, with many focused on problems like generalized linear modeling or determining the distribution(s) from which the data was drawn. For these problems the model is typically composed of the equations and distributions that are fitted to the data, which are easily expressed syntactically such that the PPL API offers an 
expressive way to define the model and extensions can be implemented in an intuitive and straightforward way.`

Line 35:

``PyAutoFit` is a PPL whose core design is providing a direct interface with the model, data, fitting procedure and 
results, allowing it to provide comprehensive management of many different aspects of model-fitting and making it 
suited to longer term software projects.. `

dfm:
Great - I think that helps to clarify things. I'm going to do one more pass of editing then we should be in good shape. Thanks!

PR #152, opened 2021-01-27
Feature/sqlite

rhayes777:
Merging at a good point - flat packing and querying model.

More work required to add metadata etc. Will also provide an example of how you can try this out on existing directory structures next week.


Jammy2211 (approved):
Not a huge amount to say, given I dont know sqlite.

I should learn as we begin to implement things.

PR #153, opened 2021-02-03
Feature/sqlite

rhayes777:
You can now try out creating a model and querying it.

```python
from autogalaxy.profiles.light_profiles import EllipticalSersic, EllipticalCoreSersic

aggregator = Aggregator.from_database(
"my_database.sqlite"
)

aggregator.add_directory("rjlens")

lens = aggregator.galaxies.lens

aggregator = aggregator.query((lens.bulge == EllipticalCoreSersic) & (lens.disk == EllipticalSersic))

samples = aggregator.values('samples')
```

The queries are on attributes of the best fit instance.


rhayes777:
The now supports .values() for obtaining samples and other pickles. Let me know how you get on

PR #162, opened 2021-03-03
feature/dataset query

rhayes777:


rhayes777:
You can now query phase_name and dataset_name. This will only work on freshly compiled databases

```python
aggregator.query(aggregator.phase_name == "something")

Jammy2211 (approved):

PR #179, opened 2021-03-31
feature/json

rhayes777:


autofit/non_linear/optimize/pyswarms.py
```diff
@@ -479,9 +480,8 @@ def sampler_fom_model_and_fitness(self, model, fitness_function, bounds, init_po
 
 class PySwarmsLocal(AbstractPySwarms):
 
-    @convert_paths
     def __init__(
-            self,
+            self, name=None,
```
Jammy2211, line 484: Put name on next line (Same for other searches).

PR #183, opened 2021-04-07
feature/database

rhayes777:


test_autofit/database/paths/test_switch.py
```diff
@@ -0,0 +1,11 @@
+import autofit as af
+
+
+def test_dynesty(session):
```
rhayes777, line 4: This shows how to switch to using a database.

You can make a session something like this:

```python 
engine = create_engine(f"sqlite:///{database_path}")
session = sessionmaker(bind=engine)()
db.Base.metadata.create_all(engine)
```

we'll need a convenience function eventually

PR #188, opened 2021-04-28
feature/identifier

rhayes777:
Customise what attributes are used to create an identifier. Only attributes of an object found in __identifier_fields__ are used if it is defined.

Currently no attributes of any search are used because an example __identifier_fields__ is defined in AbstractSearch.

This should be replaced with:

```python
__identifier_fields__ = ("include", "these", "attributes")


rhayes777:
@jonathanfrawley Presumably if we run the build for every push it doesn't need to also run for PRs?

jonathanfrawley:
Yeah, it's up to you. I think it is nice to see the status in pull requests personally so you could remove the push hook if that suits best.

rhayes777:
I'd prefer to leave push rather than pull just so you get feedback more frequently. Easier to identify a problem if it was introduced more recently

rhayes777:
The status actually shows in the PR for any checks run on that commit I think

jonathanfrawley:
Ah okay, in that case feel free to remove.

Jammy2211 (approved):

PR #190, opened 2021-04-28
feature/grid search identifier

rhayes777:


Jammy2211 (approved):

rhayes777:
A Fit is added to the database for the GridSearch, with Fits run as part of that GridSearch as children in a one to many relationship.

The aggregator can be used to query by the GridSearch or by its children. Queries performed on the instance of the GridSearch are against the median model.

```python
# Query for Fits that represent a grid search
parent_aggregator = aggregator.query(
    aggregator.is_grid_search
)
print(parent_aggregator.fits)

# Get an aggregator for children
child_aggregator = aggregator.children()
print(child_aggregator.fits)

# Query those children
print(
    child_aggregator.query(
        aggregator.centre <= 2
    ).fits
)

# Do a more complicated query for grid searches

print(
    aggregator.query(
        aggregator.is_grid_search & (aggregator.centre == 2)
    ).children().fits
)

PR #196, opened 2021-05-05
feature/identifier

rhayes777:


rhayes777:
Also allows optional saving of all samples:

```python
search = af.Search(save_all_samples=True)

Jammy2211 (approved):

PR #226, opened 2021-05-20
feature/graphical optimiser

rhayes777:
Crude implementation of optimise for classic autofit optimisers, principally expressed in classical autofit. It needs significantly more testing and probably finessing but should at least serve to illustrate how the two families of classes can be unified.

Currently will only work for factors that are defined using the declarative interface. This is so that all other autofit functionality can be easily recovered.

Updated messages (prior to projection) are created using classical autofit methods. These might not be sufficiently rigurous for this usecase.


autofit/graphical/declarative.py
```diff
@@ -130,7 +131,7 @@ def optimise(
         -------
         A collection of prior models
         """
-        self.freeze()
+        # self.freeze()
```
rhayes777, line 134: I need to figure out why this was breaking the code and reintroduce it.

autofit/non_linear/nest/abstract_nest.py
```diff
@@ -50,6 +54,81 @@ def __init__(
             **kwargs
         )
 
+    def optimise(
+            self,
+            factor: ModelFactor,
+            model_approx: EPMeanField,
+            status: Optional[Status] = None
+    ) -> Tuple[EPMeanField, Status]:
+        """
```
rhayes777, line 63: This is where the meat is.

autofit/non_linear/abstract_search.py
```diff
@@ -151,6 +153,81 @@ def __init__(
 
     __identifier_fields__ = tuple()
 
+    def optimise(
```
rhayes777, line 156: Actually moved the important stuff here - this way _any_ autofit search can now be used in conjunction with graphical

Jammy2211 (approved):

PR #254, opened 2021-06-09
identifiers account for class

### LINKED ISSUE #251: Unique id does not distinguish UniformPrior and LogUniformPrior
I think these give the same unique id:

```
gaussian_0 = af.Model(p.Gaussian)
gaussian_0.intensity = af.UniformPrior(lower_limit=1e-6, upper_limit=1e6)
```

```
gaussian_0 = af.Model(p.Gaussian)
gaussian_0.intensity = af.LogUniformPrior(lower_limit=1e-6, upper_limit=1e6)
```

Backwards compatibility is gonna be...interesting :P

rhayes777:
Fixes #251 - identifiers now account for class


Jammy2211:
Could we add a old_unique_id bool to the config and put it on this clause, so I am able to perform reruns with these current ids?

I'm not in a position to rerun a load of results currently, so this is probably the best we can do.

Jammy2211 (approved):

PR #258, opened 2021-06-09
feature/declarative global

rhayes777:
Allows analysis instances to be added:

```python
combined_analysis = analysis_1 + analysis_2
```

If general/analysis/n_cores > 1 then multiprocessing pool.map is used to parallelise evaluation.

Currently sneaky pool is not used as it swaps out arguments based on type. Here analyses would need to be distributed across processes. Will look at that now but might be 'out of scope' for this afternoon......

### DIFF
--- a/autofit/config/general.ini
+++ b/autofit/config/general.ini
@@ -13,3 +13,6 @@ ignore_prior_limits = False
 
 [test]
 test_mode = False
+
+[analysis]
+n_cores = 1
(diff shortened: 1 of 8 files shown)


Jammy2211 (approved):

Jammy2211 (approved):

rhayes777:
Now uses an implementation of a pool.

If you set analysis to use more than one core, then analyses are distributed across n_cores processes which run until the pool goes out of scope. Each instance is mapped out to each process which then calculates likelihoods for each of its analyses before passing those back in a queue to be summed by the pool.

PR #292, opened 2021-06-30
feature/lbfgs

Jammy2211:
This PR adds the scipy L-BFGS method as a PyAutoFit non linear search, including outputs, visualization, unique tags, etc.

The following points are noteworthy for this method:

- The scipy L-BFGS method does not store all parameter samples, but only the previous best estimate (`lbfgs.x`). As a result, the history of a model-fit is not accessible and its corresponding `Samples` object contains lists of length 1. We could easily add a history dictionary if we felt it was benefitial.

- The `__indentifier_fields__` which define which input parameters control the unique_tag are currently empty. It is not clear to me that any parameter makes a key difference such that results cannot be resumed.

- I have omitted jacobians, gradients, etc, for now. We should discuss how these should be implemented (most likely via the Analysis class).

- We should also discuss how this will work within the requirements on the graphical models.

An example of the autofit L-BFGS implementing of the scipy function can be found here:

`https://github.com/Jammy2211/autofit_workspace/blob/master/scripts/searches/optimize/LBFGS.py`


autofit/graphical/factor_graphs/transform.py
```diff
@@ -9,7 +9,7 @@
 from scipy.special import ndtr, ndtri
 from scipy.stats._continuous_distns import _norm_pdf
 
-# from ...mapper.operator import (
+# from ...mapper.operator import
```
rhayes777, line 12: Commented code?

autofit/non_linear/initializer.py
```diff
@@ -51,16 +50,16 @@ def from_config(cls, config):
                 lower_limit=ball_lower_limit, upper_limit=ball_upper_limit
             )
 
-    def initial_samples_from_model(self, total_points, model, fitness_function):
+    def initial_samples_from_model(self, total_points : int, model, fitness_function):
```
rhayes777, line 53: Weird formatting. Probably best over new line.

The model should be type AbstractPriorModel

autofit/non_linear/initializer.py
```diff
@@ -151,12 +150,11 @@ def __init__(self):
         The InitializerPrior class generates from the priors, by drawing all values as unit values between 0.0 and 1.0
         and mapping them to physical values via the prior.
         """
-
         super().__init__(lower_limit=0.0, upper_limit=1.0)
 
 
 class InitializerBall(Initializer):
-    def __init__(self, lower_limit, upper_limit):
+    def __init__(self, lower_limit : float, upper_limit : float):
```
rhayes777, line 157: Same here - there shouldn't be spaces before the colons

autofit/non_linear/optimize/lbfgs.py
```diff
@@ -0,0 +1,251 @@
...
+
+        for key, value in config_dict.items():
+            try:
+                config_dict[key] = self.kwargs[key]
+            except KeyError:
+                pass
+
+        return config_dict
+
+    def _fit(self, model: AbstractPriorModel, analysis : Analysis, log_likelihood_cap : float=None):
```
rhayes777, line 84: And here. Also if the default is None you should use Optional[float]

autofit/non_linear/optimize/lbfgs.py
```diff
@@ -0,0 +1,251 @@
...
+        fitness_function = self.fitness_function_from_model_and_analysis(
+            model=model, analysis=analysis
+        )
+
+        if self.paths.is_object("x0"):
+
+            x0 = self.paths.load_object("x0")
+            total_iterations = self.paths.load_object("total_iterations")
+
+            self.logger.info("Existing PySwarms samples found, resuming non-linear search.")
```
rhayes777, line 111: PySwarms?

autofit/non_linear/optimize/lbfgs.py
```diff
@@ -0,0 +1,251 @@
...
+
+        else:
+
+            initial_unit_parameter_lists, initial_parameter_lists, initial_log_posterior_list = self.initializer.initial_samples_from_model(
```
rhayes777, line 115: Oh god that's a long line. Is the initial_ prefix totally necessary?

autofit/non_linear/optimize/lbfgs.py
```diff
@@ -0,0 +1,251 @@
...
+                total_points=1,
+                model=model,
+                fitness_function=fitness_function,
+            )
+
+            x0 = np.asarray(initial_parameter_lists[0])
+
+            total_iterations = 0
+
+            self.logger.info("No PySwarms samples found, beginning new non-linear search. ")
```
rhayes777, line 125: If the logic is really similar to PySwarms might it make sense to move some of the code to a common parent class? Best not to copy and paste ;)

Why is all this PySwarms code so applicable?

autofit/non_linear/optimize/lbfgs.py
```diff
@@ -0,0 +1,251 @@
...
+
+        maxiter = self.config_dict_options["maxiter"]
```
rhayes777, line 127: You should be able to use:

```python
maxiter = self.config_dict_options.get("maxiter", 1e8)

File: autofit/non_linear/optimize/lbfgs.py [hunk 9]
```diff
@@ -0,0 +1,251 @@
...
+
+            if self.iterations_per_update > iterations_remaining:
+                iterations = iterations_remaining
+            else:
+                iterations = self.iterations_per_update
+
+            if iterations > 0:
+
+                config_dict_options = self.config_dict_options
+                config_dict_options["maxiter"] = iterations
```
rhayes777, line 144: Updating the config here looks weird and is probably not the best way to do things

autofit/non_linear/optimize/lbfgs.py
```diff
@@ -0,0 +1,251 @@
...
+        maxiter = self.config_dict_options["maxiter"]
+
+        if maxiter is None:
+            maxiter = 1e8
+
+        while total_iterations < maxiter:
+
+            iterations_remaining = maxiter - total_iterations
+
+            if self.iterations_per_update > iterations_remaining:
```
rhayes777, line 136:
```python
iterations = min(self.iterations_per_update, iterations_remaining)

File: autofit/non_linear/optimize/lbfgs.py [hunk 11]
```diff
@@ -0,0 +1,251 @@
...
+    @property
+    def samples(self):
+        """
+        Create a `Samples` object from this non-linear search's output files on the hard-disk and model.
+
+        For L-BFGS, only a single ndarray of parameters is returned, alongside its log posterior value. Thus, the
+        `LBFGSSamples` object contains only one entry.
+        """
+
+        if self._samples is not None:
```
rhayes777, line 233: You can get rid of one whole line of code by checking inverting this logic and using only one return statement!

autofit/non_linear/paths/directory.py
```diff
@@ -156,7 +156,7 @@ def save_all(self, search_config_dict, info, pickle_files):
     @make_path
     def _pickle_path(self) -> str:
         """
-        This is private for a reason - use the save_object etc. methods to save and load pickles
+        This is private for a reason, use the save_object etc. methods to save and load pickles
```
rhayes777, line 159: You can make it _really_ private with a double underscore

test_autofit/non_linear/optimize/test_lbfgs.py
```diff
@@ -0,0 +1,96 @@
...
+        assert lbfgs.config_dict_options["iprint"] == -1
+        assert lbfgs.config_dict_options["maxls"] == 20
+        assert lbfgs.config_dict_options["maxiter"] == 15000
+        assert lbfgs.config_dict_options["disp"] == False
+        assert isinstance(lbfgs.initializer, af.InitializerPrior)
+        assert lbfgs.iterations_per_update == 11
+
+    def test__samples_from_model(self):
+        
+        LBFGS = af.LBFGS()
```
rhayes777, line 70: Variable names should be lowercase

rhayes777: If I could do a vomit emoji I would

test_autofit/output/non_linear/LBFGS/tag/pickles/total_iterations.pickle
```diff
@@ -0,0 +1 @@
+�K.
```
rhayes777, line 1: Are pickles necessary? Might make sense to gitignore them as they change with runs adding crap to the git history

rhayes777 (approved):

autofit/non_linear/optimize/lbfgs.py
Jammy2211: I will probably attempt a refactor in this direction in the future. The problem is every optimizer has its own API and intrincacies and trying to generalize everything into one normally leads to very hard to catch bugs.

autofit/non_linear/paths/directory.py
Jammy2211: You wrote that code and comment lol

test_autofit/non_linear/optimize/test_lbfgs.py
Jammy2211: 🤮

test_autofit/output/non_linear/LBFGS/tag/pickles/total_iterations.pickle
Jammy2211: They are for the unit test, would be hard to omit them without running a full search to create them.

PR #308, opened 2021-07-08
Bugfix/find mode

FiachAntaw:
The implementation of `.from_mode` in TransformedMessage should call the `.from_mode` function of the underlying message class to correctly set up the message, but instead calls itself - resulting in an infinite loop and eventual interpreter crash. This patch changes the function to dispatch to the correct class method, and adds some special-case code to correctly handle nested transforms.

Note that Message transformations are currently implemented by subclassing the original message class and, if the original message class does not inherit from it, mixing in the TransformedMessage class. Because of this, the `super(TransformedMessage, cls).from_mode` call will skip intermediate transforms and dispatch directly to the original message class - and so nested TransformedMessages must be handled explicitly.

Thoughts? Happy to make changes if needed.

### DIFF
--- a/autofit/graphical/messages/abstract.py
+++ b/autofit/graphical/messages/abstract.py
@@ -770,6 +770,18 @@ class TransformedMessage(AbstractMessage):
     # TODO add code for analytic hessians when Jacobian is fixed e.g. for shifted messages
     logpdf_gradient_hessian = AbstractMessage.numerical_logpdf_gradient_hessian
 
+    @staticmethod
+    def _from_mode(
+            cls: "TransformedMessage",
+            mode: np.ndarray,
+            covariance: np.ndarray
+    ) -> "AbstractMessage":
+        mode, jac = cls._transform.transform_jac(mode)
+        covariance = jac.invquad(covariance)
+        if cls._depth > 1:
+            return TransformedMessage._from_mode(cls._Message, mode, covariance)
+        return mode, covariance
+
     @classmethod
     def from_mode(
             cls,
@@ -777,6 +789,5 @@ class TransformedMessage(AbstractMessage):
             covariance: np.ndarray,
             id_=None
     ) -> "AbstractMessage":
-        mode, jac = cls._transform.transform_jac(mode)
-        covariance = jac.invquad(covariance)
-        return cls._Message.from_mode(mode, covariance, id_=id_)
+        mode, covariance = TransformedMessage._from_mode(cls, mode, covariance)
+        return super(TransformedMessage, cls).from_mode(mode, covariance, id_=id_)
--- a/autofit/graphical/messages/abstract.py
+++ b/autofit/graphical/messages/abstract.py
@@ -770,6 +770,18 @@ class TransformedMessage(AbstractMessage):
     # TODO add code for analytic hessians when Jacobian is fixed e.g. for shifted messages
     logpdf_gradient_hessian = AbstractMessage.numerical_logpdf_gradient_hessian
 
+    @staticmethod
+    def _from_mode(
+            cls: "TransformedMessage",
+            mode: np.ndarray,
+            covariance: np.ndarray
+    ) -> "AbstractMessage":
+        mode, jac = cls._transform.transform_jac(mode)
+        covariance = jac.invquad(covariance)
+        if cls._depth > 1:
+            return TransformedMessage._from_mode(cls._Message, mode, covariance)
+        return mode, covariance
+
     @classmethod
     def from_mode(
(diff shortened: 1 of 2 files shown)


matthewghgriffiths:
Good catch, I think a simpler bugfix would be to change

`return cls.from_mode(mode, covariance, id_=id_)`

to 

`return cls._Message.from_mode(mode, covariance, id_=id_)`

In `TransformedMessage.from_mode`

This should recursively call the parent transformed messages correctly

FiachAntaw:
That was my first thought, too, but `cls._Message.from_mode` will return an object of the untransformed message type (i.e. `LogNormalMessage.from_mode` returns a bare `NormalMessage` object). Since .from_mode is a classobject, it's bound to the class that it belongs to unless called via `super()`. I've changed the test case to verify the returned message type.

Another approach is to roll .from_mode into the dynamically generated `Transformed` class. Then the traditional `super()` dispatch approach will work correctly, removing the need to special-case nested transforms. Maybe that's a little cleaner?

Looking at the code again, I'm wondering what happens if a user sub-classes a transformed message object and then transforms their sub-class. Because the TransformedMessage class is only inherited by the first set of transforms, the user subclass methods will override the methods on the final transform. i.e.

```python
import autofit.graphical as graph

class LogNormalMessage2(graph.LogNormalMessage):
        @classmethod
        def from_mode(cls, mode, covariance, id_=None):
            # Should be called with mode of 1
            assert mode == 1
            return super(LogNormalMessage2, cls).from_mode(mode, covariance, id_=id_)
LN2Shifted = LogNormalMessage2.shifted(shift=-1)
m = LN2Shifted.from_mode(0, 10)
```

Fails because the call to `LN2Shifted.from_mode` dispatches to `LogNormalMessage2.from_mode` before `TransformedMessage.from_mode`. The same issue should affect the other methods, too. I'm not sure what the best solution for this would be, other than to change from a mix-in pattern to some other means of generating transformed classes.

Jammy2211:
Every looks OK to me, but this is more Rich's domain. He's working next week, so will give this a proper look then (he had a quick look today but figured it wasn't a 5 minute job).

Once he's looked we'll definitely merge this into the source code though!

matthewghgriffiths:
I've had another go at fixing the bug, I've also added another test case to test out some more complicated transforms.

autofit/graphical/messages/abstract.py
```diff
@@ -777,6 +777,8 @@ def from_mode(
             covariance: np.ndarray,
             id_=None
     ) -> "AbstractMessage":
-        mode, jac = cls._transform.transform_jac(mode)
-        covariance = jac.invquad(covariance)
-        return cls.from_mode(mode, covariance, id_=id_)
+        mode, covariance = TransformedMessage._from_mode(cls, mode, covariance)
```
rhayes777, line 780: This line is weird broken.

```
AttributeError: type object 'TransformedMessage' has no attribute '_from_mode'
```

It also looks a bit weird anyway?

rhayes777:
I'd suggest we avoid the dynamic generation of classes altogether. Yes it's very clever, but if we can implement the same logic without it that will make it much easier to avoid these kind of bugs.

matthewghgriffiths:
I think that's a fair point. 

The `AbstractMessage` was set up so you could potentially specify a mean-field declaratively as a class structure without having to specify concrete parameters for the Messages, however in the current API I don't think we rely on the more declarative style, and any time we deal with an `AbstractMessage` we only deal with concrete, parametrised instances of them.

Does autofit only use concrete instances of Priors? In that case it would probably be straightforward to set almost all the static/classmethods of AbstractMessages to standard methods. The transformation API would have be altered, but that shouldn't be too difficult.

rhayes777:
I'm not against having an AbstractClass. We have an abstract Prior class simply called "Prior". It's pretty conventional and useful!

It's more the TransformedMessage code I find confusing to read. Defining new classes on the fly during execution adds (probably) unnecessary complexity.

matthewghgriffiths:
what I mean is that atm you could define a class like:

```
class MeanFieldModel(AbstractMeanFieldModel):
    x = NormalMessage
    y = NormalMessage
    z = NormalMessage.transformed(log_transform)
```

And it would be relatively straightforward to parse this model directly into EP, however, I don't think we need this capability, or are likely to need this use case?

rhayes777:
Yeah it seems kind of unnecessary when you can just pass kwarg messages into a constructor

FiachAntaw:
Sounds good to me, I'm happy to close this PR in favour of refactoring the API.

PR #318, opened 2021-07-21
feature/grid identifier

### LINKED ISSUE #317: Remove number_of_cores from GridSearch unique identifer
.

rhayes777:
It is now possible to exclude rather than include fields for an identifier

```python
__exclude_identifier_fields__ = ("number_of_cores",)
```

If `__identifier_fields__` is defined this takes precedence.

Resolves #317


Jammy2211 (approved):

Jammy2211:
Does `result_output_interval`, which we excluded from the GridSearch identifier last week, use this new API?

rhayes777:
Any attributes that start with an underscore are also excluded. To be fair that might have been a better solution to this problem heh

rhayes777:
I guess explicit is better than implicit...

PR #320, opened 2021-07-22
feature/mapper by path

rhayes777:
Recovery of model instances from samples relied on prior ordering. Now it uses the name of the prior. This is still a bit dodgy but backwards compatible.

That is, a prior with path model.sub_model.prior gets a name model_sub_model_prior in samples. This is matched back to the path. The risk is that (in theory) two priors could have the same effective name. However, this would have been an issue regardless.

The better solution is to change kwargs to use paths instead of strings. I'm concerned this will break backwards compatibility but I'm going to try it...

### DIFF
--- a/autofit/mapper/prior_model/abstract.py
+++ b/autofit/mapper/prior_model/abstract.py
@@ -796,12 +796,29 @@ class AbstractPriorModel(AbstractModel):
             arguments
         )
 
+    def path_for_name(self, name):
+        exploded = tuple(name.split("_"))
+        for path, _ in self.path_priors_tuples:
+            exploded_path = tuple(
+                string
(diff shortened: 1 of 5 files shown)


rhayes777:
Ok integrated and tested - it seems to respect backwards compatability and eliminate the issue posed in the example.py

rhayes777:
It uses the best possible method available - for backwards compatability it will match to names, but in future the keys in the sample.kwargs dictionary are paths. This is much safer. Still warrants testing obviously.

Jammy2211:
I look forward to breaking this.

rhayes777:
:'(

Jammy2211 (approved):

PR #322, opened 2021-07-22
feature/prior match

rhayes777:
Attempted to replace loading dynesty and creating samples with simply loading samples to mitigate error caused by mismatch between order of priors between models. Also fixed issue recovering tuple values from their names.

- prior_id/example.py runs now (well it's still running...)
- I haven't implemented this for other optimisers yet although the pattern should be obvious
- I'd recommend testing carefully before merging this branch


autofit/non_linear/samples.py
```diff
@@ -152,8 +152,7 @@ def from_lists(
         """
         samples = list()
 
-        # Another speed up.
-        paths = model.paths
+        paths = model.unique_prior_paths
```
rhayes777, line 155: @Jammy2211 ffs...

Jammy2211 (approved):

Jammy2211 (approved):

PR #324, opened 2021-07-24
Feature/covariance

Jammy2211:


autofit/non_linear/samples.py
```diff
@@ -877,6 +888,23 @@ def offset_vector_from_input_vector(self, input_vector : List) -> [float]:
...
+
+        Follow that link for a description of what the covariance matrix is.
+
+        Returns
+        -------
+        ndarray
+            A covariance matrix of shape [total_parameters, total_parameters] for the model parameters of the
+            non-linear search.
+        """
+        return np.cov(m=self.parameter_lists, rowvar=False, aweights=self.weight_list)
```
matthewghgriffiths, line 906: As I'm not a proper statistician I don't know which one is correct, but you might want to pass `bias=True` to `np.cov` to calculate the covariance (at the very least I think you may want to have `bias=True/False` as a keyword argument for `covariance_matrix`

PR #327, opened 2021-08-04
feature/grid result

### LINKED ISSUE #326: Grid Search Results
I am working on the following branch:

https://github.com/rhayes777/PyAutoFit/tree/feature/subhalo_result

I have two requests for extending the GridSearchResult:

**Information on the physical values of parameters for every grid cell:**

The inputs `lower_limit_lists` and `physical_lower_limits_lists` contain information on the unit prior values and physical values of every parameter on the grid search. The physical value of every prior and their upper limits are computed via the following properties:

```
    @property
    def physical_step_sizes(self) -> Tuple[int]:
        """
        The largest physical step sizes of every parameter

        Returns
        -------

        """

        physical_step_sizes = []

        # TODO : Make this work for all dimensions in a less ugly way.

        for dim in range(self.no_dimensions):

            values = [value[dim] for value in self.physical_lower_limits_lists]
            diff = [abs(values[n] - values[n - 1]) for n in range(1, len(values))]

            if dim == 0:
                physical_step_sizes.append(np.max(diff))
            elif dim == 1:
                physical_step_sizes.append(np.min(diff))
            else:
                raise exc.GridSearchException(
                    "This feature does not support > 2 dimensions"
                )

        return tuple(physical_step_sizes)

    @property
    def physical_centres_lists(self):
        return [
            [
                lower_limit[dim] + self.physical_step_sizes[dim] / 2
                for dim in range(self.no_dimensions)
            ]
            for lower_limit in self.physical_lower_limits_lists
        ]

    @property
    def physical_upper_limits_lists(self):
        return [
            [
                lower_limit[dim] + self.physical_step_sizes[dim]
                for dim in range(self.no_dimensions)
            ]
            for lower_limit in self.physical_lower_limits_lists
        ]
```

However, these only work for uniform priors and therefore do not provide sufficient informaiton to determine the physical value of every prior of the grid search if a LogUniformPrior is assumed. Can you make it so these values are fully accessible for the general use case:

**Mapping Results back to the native shape of the grid:**

The GridSearchResult condenses all the information about the result into 1D lists. For example, the results of a 2x2 grid search are stored as a list with 4 entries and the `lower_limit_lists` is stored as a list of list, where every entry of the list contains the unit priors of a grid cell.

For example if there are two dimensions on a grid search of shape 2x2:

lower_limit_lists = [[0.0, 0.0], [0.0, 0.5], [0.5, 0.0], [0.5, 0.5]].

I want to make sure that we can map these results back to a grid search of its native dimesions (E.g. from a 1D list to a numpy array or list of lists of shape 2x2). The following method does this:

```
    @property
    def results_native(self):
        """
        The result of every grid search on a NumPy array whose shape is the native dimensions of the grid search.

        For example, for a 2x2 grid search the shape of the Numpy array is (2,2) and it is numerically ordered such
        that the first search's result (corresponding to unit priors (0.0, 0.0)) are in the first value (E.g. entry
        [0, 0]) of the NumPy array.
        """
        return np.reshape(
            np.array([result for result in self.results]),
            self.shape_native,
        )
```

Can you put some unit tests and check that this will work for higher dimensional grids? I am confident it works for 2x2, but am unsure about higher dimensionality.

rhayes777:
closes #326


Jammy2211 (approved):

rhayes777:
Doesn't deal with physical centres. At the moment that function returns a tuple - it's not clear how this should be adapted for the non-uniform case

PR #333, opened 2021-08-06
feature/unification

rhayes777:
Intermediate step on the way to unification.

A bit messy, but importantly NormalMessages and GaussianPriors are now essentially the same thing.

Also, if a prior or a message is derived from another prior or message it retains the same identifier. This simplifies optimisation with classic autofit optimisers. I think it will ultimately lead to a good deal of simplification with some of the factor/meanfield/message manipulation and may allow for a meanfield and a collectionpriormodel to be unified.


autofit/graphical/factor_graphs/jacobians.py
```diff
@@ -1,18 +1,24 @@
...
+    Tuple, Dict, Callable, Optional, Union
+)
 
 import numpy as np
 
-from autofit.mapper.variable import Variable
 from autofit.graphical.factor_graphs.abstract import \
+    (
     FactorValue, JacobianValue
+)
```
Jammy2211, line 12: This is ugly

Jammy2211 (approved):

PR #339, opened 2021-08-18
feature/deference

rhayes777:
Initial implementation of deferred instantiation of transforms.

Transformed classes now track and instantiate a wrapped class rather than depending on dual inheritance and wrapped class

Needs some more work - I wanted to be sure I hadn't broken too much at this point

### DIFF
--- a/autofit/database/model/model.py
+++ b/autofit/database/model/model.py
@@ -7,7 +7,6 @@ from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import relationship
 
 from autofit.tools.util import get_class_path, get_class
-import numpy as np
 
 Base = declarative_base()
 
--- a/autofit/database/model/model.py
+++ b/autofit/database/model/model.py
@@ -7,6 +7,7 @@ from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import relationship
 
 from autofit.tools.util import get_class_path, get_class
+import numpy as np
 
 Base = declarative_base()
 
@@ -97,12 +98,15 @@ class Object(Base):
         """
 
         from autofit.mapper.prior_model.prior_model import PriorModel
-        from autofit.mapper.prior.prior import Prior
+        from autofit.mapper.prior.abstract import Prior
         from autofit.mapper.prior_model.collection import CollectionPriorModel
 
         if source is None or isinstance(
                 source,
-                np.ndarray
+                (
+                        np.ndarray,
+                        np.broadcast
+                )
         ):
             from .instance import NoneInstance
             instance = NoneInstance()
--- a/autofit/messages/transformed.py
+++ b/autofit/messages/transformed.py
@@ -124,10 +124,9 @@ class TransformedMessage(AbstractMessage):
         # noinspection PyUnresolvedReferences
         return self._transform.inv_transform(self._Message.variance.func(self))
 
-    @classmethod
-    def _sample(cls, self, n_samples) -> np.ndarray:
-        x = cls._Message._sample(self, n_samples)
-        return cls._transform.inv_transform(x)
+    def _sample(self, n_samples) -> np.ndarray:
+        x = self.instance._sample(n_samples)
+        return self._transform.inv_transform(x)
 
     @classmethod
     def _logpdf_gradient(  # type: ignore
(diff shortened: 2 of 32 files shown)


rhayes777:
Still not sure this works as some class methods that depend on transform may not be convertible because they are used within the context of an instance

autofit/messages/transformed.py
```diff
@@ -0,0 +1,171 @@
...
+from .transform import AbstractDensityTransform
+from ..tools.cached_property import cached_property
+
+
+class TransformedMessage(AbstractMessage):
+    _Message: Type[AbstractMessage]
+    _transform: Union[AbstractDensityTransform, Type[AbstractDensityTransform]]
+    _depth = 0
+
+    def __init__(self, *args, **kwargs):
```
matthewghgriffiths, line 16: It might be nice/better/easier to use `__new__` to create a new class with the _transform attributed initialised, but it's probably less work to fix this version of the implementation.

test_autofit/graphical/test_unification.py
```diff
@@ -38,3 +41,127 @@ def test_bad_id(
...
+        upper_limit=upper_limit,
+    ).value_for(
+        unit_value
+    ) == pytest.approx(
+        physical_value
+    )
+
+
+def test_uniform_odd_result():
+    prior = af.UniformPrior(16, 32)
```
rhayes777, line 164: This gives me inf rather than 16 - I guess because the transformed distribution is not well behaved in that domain?

@matthewghgriffiths

autofit/messages/transformed.py
matthewghgriffiths: You could also define two classes, 
```
class UninitialisedTransformedMessage(TransformedMessage):
     _Message: Type[AbstractMessage]
     _transform: Type[AbstractDensityTransform]
     
     def __new__(cls, *args, **kwargs):
           transform_args = inspect.getfullargspec(
                self._transform
            ).args[1:]

            transform_dict = dict()
            for arg in transform_args:
                if arg in kwargs:
                    transform_dict[
                        arg
                    ] = kwargs.pop(arg)
            transform = cls._transform(
                **transform_dict
            )
            return cls._Message.transformed(transform)
```

rhayes777: Hmm that might be nicer. One of my concerns at the moment is the inconsistency with which I've converted classmethods. Some will just fail for bad reasons. Will consider trying this

test_autofit/graphical/test_unification.py
matthewghgriffiths: is the `value_for` method returning the log probability density of the prior?

rhayes777: It maps from a value in the range 0 -> 1 to a physical value. So sort of the inverse of that?

matthewghgriffiths: So I think it's the percent point function (`ppf`) as used by `scipy.stats`?

I think the challenge is that for the normal distribution ppf(0) = -inf and ppf(1) = inf

matthewghgriffiths: So I think for these transformed values work for the open interval (0, 1), but not for 0 or 1

Have you tried values between 0 and 1?

rhayes777: it seems to still work at 0 or 1 and up to ~10, at least in terms of the limits as defined for the UniformPrior.

Those limits are used to compute shift and scale

rhayes777: Might be worth a call if you have time

autofit/database/model/model.py
```diff
@@ -7,6 +7,7 @@
 from sqlalchemy.orm import relationship
 
 from autofit.tools.util import get_class_path, get_class
+import numpy as np
```
Jammy2211, line 10: Import numpy above autofit imports

Jammy2211 (approved):

rhayes777:
I couldn't get the UninitialisedTransformedMessage idea to work but will look again next week.

All Priors are now really Messages. I still need to test what happens when I put them into the declarative framework.

Will merge as it stands because I don't think it breaks anything. Definitely worth testing functionality and efficiency of classical autofit given these changes.

PR #340, opened 2021-08-18
Feature/modify before

Jammy2211:


autofit/non_linear/abstract_search.py
```diff
@@ -445,6 +445,8 @@ class represented by model M and gives a score for their fitness.
         self.paths.unique_tag = self.unique_tag
         self.paths.restore()
 
+        analysis.modify_before_fit(model=model)
```
rhayes777, line 448: Might be worth making this return a new analysis rather than modifying the existing one. Otherwise when you reuse the analysis for different searches you might get unexpected results

autofit/non_linear/analysis/analysis.py
```diff
@@ -33,6 +34,15 @@ def save_results_for_aggregator(self, paths: AbstractPaths, model: CollectionPri
                                     samples: OptimizerSamples):
         pass
 
+    def modify_before_fit(self, model: AbstractPriorModel):
```
rhayes777, line 37: Here you'd need to return self

autofit/non_linear/analysis/analysis.py
```diff
@@ -33,6 +34,15 @@ def save_results_for_aggregator(self, paths: AbstractPaths, model: CollectionPri
...
+        """
+        Overwrote this method to modify the attributes of the `Analysis` class before the non-linear search begins.
```
rhayes777, line 39: Overwrite

rhayes777 (approved):

autofit/non_linear/abstract_search.py
Jammy2211: Do you mean a copy of the analysis, or to simple use return analysis?

autofit/non_linear/analysis/analysis.py
Jammy2211: Ah.

PR #345, opened 2021-08-25
feature/alternate transform

rhayes777:
- Removes defered transformations
- Instead, the new method of priors is used to generate a transformed class on the fly. This may cause issues with typing or database but tests do pass.
- Added tests demonstrating that UniformPriors can now be used with the graphical framework.
- Reverted many methods back to classmethods
- Changes mu to mean in NormalMessage to support BC
- Used __identifier_fields__ to prevent transform/child messages appearing in identifiers


rhayes777:
Still trying to figure out what the arithmetic should do. The second assertion fails because of a nan?

```python

def test_prior_arithmetic():
    prior = af.UniformPrior(
        lower_limit=10,
        upper_limit=20
    )

    divided = prior / prior
    assert divided.value_for(0.5) != prior.value_for(0.5)
    assert (divided * prior).value_for(0.5) == prior.value_for(0.5)

Jammy2211 (approved):

PR #347, opened 2021-09-01
feature/transform wrapper no inherit

rhayes777:
This provides a working solution for priors which are really messages that have transforms generated on the fly.

All the tests pass but it really needs more thought. I'm still concerned all I'm doing is piling hack on top of hack and there ought to be a more elegant solution.

The principal problem is how transformations work. A new class is generated for each transform that extends the original class. With a bit of extra magic it was possible to make this pickle ok. However, this no longer works when new classes are generated on the fly.

For a uniform prior we want to create a transform on-the-fly that shifts a uniform normal distribution to some specific range. We can do this, but then when we try to pickle the definition of this new class we hit a problem. If it's not instantiated we can't use any tricks like reduce or getstate because it's a class not an instance. We could set the location that pickle will try to locate it and then implement some custom getter behaviour, but then we would need to be able to instantiate a new version on the fly.

The point is that pickle wants to find a class somewhere. If that class doesn't exist it has to be made. If the class has some arbitrary configuration then somehow pickle has to communicate that.

Anyway, most solutions didn't work so I've ended up wrapping the transformed 'class' and 'instance' in instances of another class that can throw away and recreate the transformed definitions on the fly. Through a lot of magic I've made prior play nicely with this. However, I'm really not convinced this is a nice way to do things.

Feel free to look at and test this solution and if it works you can use it. Otherwise I think we need to consider throwing out the notion of defining transforms as on-the-fly classes and refactor the message class to permit transformation purely by composition of instances.

### DIFF
--- a/test_autofit/graphical/regression/test_identifier.py
+++ b/test_autofit/graphical/regression/test_identifier.py
@@ -118,7 +118,7 @@ def make_model():
 def test_model_identifier(
         model
 ):
-    assert str(model) == "1719f29d2938d146d230d52ef7379a84"
+    assert str(model) == "626c1beb5b1fb99781f30950595f36de"
 
 
 def test_model_identifier_fields(
@@ -134,7 +134,7 @@ def test_model_identifier_fields(
         '0.0',
         'upper_limit',
         '1.0',
-        'normalization',
+        'intensity',
         'UniformPrior',
         'lower_limit',
         '0.0',


rhayes777:
Looking at the diff makes it look like so little work! It might be ok but basic things like checking if a UniformPrior is an instance of a UniformPrior will no longer work correctly.

rhayes777:
I've actually found a good solution. Now Priors are instances of the transformed instance wrapper meaning they can hold a reference to and manipulate a message without too much magic.

Jammy2211:
Because Rich is on holiday and messages have proven extremely troublesome in breaking backwards compatibiltiy, I have reverted the master branch to its state August 10th. 

This may cause issues merging this PR (which we don't want to do for a few weeks when Rich is back AND once we've given a bit more thought to things).

However, it currently looks like the revert has worked OK...

Jammy2211:
Ok, I have now properly reverted master and conflicts are here!

There is a `master_backup` branch... this is all fun stuff we can figure out once Rich is back. Hopefully I wont touch master much for a while.

PR #349, opened 2021-09-23
attempt to fix deep recursion with from_mode class method giving eventual nan

rhayes777:


rhayes777:
I realised that the from_mode function needed porting over to the new transform wrappers. However, I'm now getting nan after a few optimisations. Any ideas @matthewghgriffiths ?

rhayes777:
Added some docs explaining how all this stuff works...

Jammy2211 (approved):

PR #350, opened 2021-09-23
feature/no output

rhayes777:


Jammy2211 (approved):

aadhikaarika-lab:
Vendored and generated code is excluded
Empty files and files over 350 kiB are excluded
Only UTF-8 encoded files are indexed
Very large repositories may not be indexed

PR #393, opened 2021-11-03
feature/eden ast

rhayes777:
Improves edenisation by using ast instead of regex to parse Python.

- Fixes issue where only partial files would be parsed due edge cases
- Properly handles imports from dependency projects. e.g. autoconf packages will be correctly prefixed
- Simplifies implementation making it easier to understand
- Replaces aliased imports with explicit imports

Outstanding issues:
- It can't handle importing aliases from another file. So if a class I is aliased as A in file a and then imported into file b it won't convert properly.
- Formatting is a bit whack. Code will run but astunparse behaves a bit oddly. Need to experiment with python 3.8's inbuilt astunparse


rhayes777:
*cough*

Jammy2211:
docstrings also have issues, whereby enters are being replaced by \n and appear all formatted on a single line:

```
       '\n        Abstract base class for non-linear searches.L\n\n        This class sets up the file structure for the non-linear search, which are standardized across all non-linear\n        searches.\n\n        Parameters\n        ------------\n        name\n            The name of the search, controlling the last folder results are output.\n        path_prefix\n            The path of folders prefixing the name folder where results are output.\n        unique_tag\n            The name of a unique tag for this model-fit, which will be given a unique entry in the sqlite database\n            and also acts as the folder after the path prefix and before the search name.\n        prior_passer\n            Controls how priors are passed from the results of this `NonLinearSearch` to a subsequent non-linear search.\n        initializer\n            Generates the initialize samples of non-linear parameter space (see autofit.non_linear.initializer).\n        session\n            An SQLAlchemy session instance so the results of the model-fit are written to an SQLite database.\n        '
```

Jammy2211:
Code spacing in functions is also lost not that its a huge deal:

```
    def fit(
        self,
        model,
        analysis: "Analysis",
        info=None,
        pickle_files=None,
        log_likelihood_cap=None,
    ) -> Union[("Result", List["Result"])]:
        "\n        Fit a model, M with some function f that takes instances of the\n        class represented by model M and gives a score for their fitness.\n\n        A model which represents possible instances with some dimensionality is fit.\n\n        The analysis provides two functions. One visualises an instance of a model and the\n        other scores an instance based on how well it fits some data. The search\n        produces instances of the model by picking points in an N dimensional space.\n\n        Parameters\n        ----------\n        log_likelihood_cap\n        analysis : af.Analysis\n            An object that encapsulates the data and a log likelihood function.\n        model : ModelMapper\n            An object that represents possible instances of some model with a\n            given dimensionality which is the number of free dimensions of the\n            model.\n        info : dict\n            Optional dictionary containing information about the fit that can be loaded by the aggregator.\n        pickle_files : [str]\n            Optional list of strings specifying the path and filename of .pickle files, that are copied to each\n            model-fits pickles folder so they are accessible via the Aggregator.\n\n        Returns\n        -------\n        An object encapsulating how well the model fit the data, the best fit instance\n        and an updated model with free parameters updated to represent beliefs\n        produced by this fit.\n        "
        self.logger.info("Starting search")
        self.paths.model = model
        self.paths.unique_tag = self.unique_tag
        self.paths.restore()
        analysis = analysis.modify_before_fit(paths=self.paths, model=model)
        if (not self.paths.is_complete) or self.force_pickle_overwrite:
            self.logger.info("Saving path info")
            self.paths.save_all(
                search_config_dict=self.config_dict_search,
                info=info,
                pickle_files=pickle_files,
            )
            analysis.save_attributes_for_aggregator(paths=self.paths)
        if not self.paths.is_complete:
            self.logger.info("Not complete. Starting non-linear search.")
            self.timer.start()
            self._fit(
                model=model, analysis=analysis, log_likelihood_cap=log_likelihood_cap
            )
            self.paths.completed()
            samples = self.perform_update(
                model=model, analysis=analysis, during_analysis=False
            )
            analysis.save_results_for_aggregator(
                paths=self.paths, model=model, samples=samples
            )
            self.paths.save_object("samples", samples)
        else:
            self.logger.info(f"Already completed, skipping non-linear search.")
            try:
                samples = self.paths.load_object("samples")
            except Exception:
                samples = self.samples_via_results_from(model=model)
            if self.force_pickle_overwrite:
                self.logger.info("Forcing pickle overwrite")
                self.paths.save_object("samples", samples)
                self.paths.save_object("results", samples.results)
                analysis.save_results_for_aggregator(
                    paths=self.paths, model=model, samples=samples
                )
        result = analysis.make_result(samples=samples, model=model, search=self)
        analysis = analysis.modify_after_fit(
            paths=self.paths, model=model, result=result
        )
        self.logger.info("Removing zip file")
        self.paths.zip_remove()
        return result
```

Jammy2211:
For a reason I don't understand the following import in `autofit/non_linear/paths/sub_directory_paths.py` disappears after edenising:

`from autofit.non_linear.paths.database import DatabasePaths`

The other nearly identical imports around it are fine:

```
from VIS_CTI_Autofit.VIS_CTI_NonLinear.VIS_CTI_Paths.abstract import AbstractPaths

from VIS_CTI_Autofit.VIS_CTI_NonLinear.VIS_CTI_Paths.directory import DirectoryPaths
```

Jammy2211:
Overall though this seems way better than the old method and massively simplifies the conversion procedure!

rhayes777:
Wait so aliasing imports is fine provided that they're not our dependencies? That's madness.

Is it really a problem if we don't use a numpy alias?

rhayes777:
I think all the issues you've raised should now be fixed except for code spacing

PR #399, opened 2021-11-10
feature/mp directories

### LINKED ISSUE #390: Graphical Model w/ EP Output Structure
The following script runs a simple graphical model using EP:

https://github.com/Jammy2211/autofit_workspace_test/blob/main/graphical/expectation_propagation.py

The output of the results is as follows:

image

Currently, it is unclear what result corresponds to what factor, and what factors were refit, and the order of those refits.

**Folder Structure**

We should add an input variable `folder` to `AnalysisFactor` that add the folder name to its output path, for a more concise API (we should not use the `unique_tag` of a search, as this will create nasty conflicts with paths):

```
dynesty = af.DynestyStatic(
    path_prefix=path.join("graphical"),
    name="expectation_propagation",
    nlive=100,
    sample="rwalk",
)

analysis_factor_0 = g.AnalysisFactor(
    prior_model=prior_model_0, analysis=analysis_0, optimiser=dynesty, folder="dataset_name_0"
)
analysis_factor_1 = g.AnalysisFactor(
    prior_model=prior_model_1, analysis=analysis_1, optimiser=dynesty, folder="dataset_name_1"
)
analysis_factor_2 = g.AnalysisFactor(
    prior_model=prior_model_2, analysis=analysis_2, optimiser=dynesty, folder="dataset_name_2"
)
```

Now each `AnalysisFactor` has a `folder` we can think about how to structure the output results. First, I think each set of results should be in the folder corresponding to its `folder=dataset_name_#` (based on how paths work I think this is actually what will happens as soon as the search :

```
expectation_propagation
   -> dataset_name_0
   -> dataset_name_1
   -> dataset_name_2
```

If the AnalyisFactor's search does not have a `folder` , we should instead use an index with some generic folder names:

```
expectation_propagation
   -> factor_0
   -> factor_1
   -> factor_2
```

**Factor Optimization Order**

After adding folders, it is still not clear which unique tag corresponds to which optimiztion of the factor

```
expectation_propagation
   -> factor_0
         sdijdfiogjioudfnguidfshgfsd
         sdlkf;dlf;aktojopjfophgkokopf
   -> factor_1
        ncjkncjknckjnjkfnkjfnkjfnfkj
        sdpodpslfokfdospkfopkds
   -> factor_2
      dsoapsomdspomfopsdd
      dsfniodngfioujhoifdsj
```

We have two options, eiither add another folder expressing this:

```
expectation_propagation
   -> factor_0
      ->   optimize_0
                   sdijdfiogjioudfnguidfshgfsd
      -> optimize_1
                  sdlkf;dlf;aktojopjfophgkokopf
   -> factor_1
      ->   optimize_0
                ncjkncjknckjnjkfnkjfnkjfnfkj
      ->   optimize_1
                  sdpodpslfokfdospkfopkds
   -> factor_2
      ->   optimize_0
               dsoapsomdspomfopsdd
      ->   optimize_1
             dsfniodngfioujhoifdsj
```

Or append the unique tags with this index for conciseness (as the risk of making database interfacing more complicated)

```
expectation_propagation
   -> factor_0
         0_sdijdfiogjioudfnguidfshgfsd
         1_sdlkf;dlf;aktojopjfophgkokopf
   -> factor_1
        0_ncjkncjknckjnjkfnkjfnkjfnfkj
        1_sdpodpslfokfdospkfopkds
   -> factor_2
      0_dsoapsomdspomfopsdd
      1_dsfniodngfioujhoifdsj
```

rhayes777:
Closes #390

### DIFF
--- a/autofit/graphical/mean_field.py
+++ b/autofit/graphical/mean_field.py
@@ -1,3 +1,4 @@
+import logging
 from functools import reduce
 from typing import (
     Dict, Tuple, Optional, List, Union, Iterable
@@ -5,6 +6,7 @@ from typing import (
 
 import numpy as np
 
+from autofit import exc
 from autofit.graphical.factor_graphs import (
     Factor, AbstractNode, FactorValue, JacobianValue
 )
@@ -19,6 +21,10 @@ from autofit.messages.fixed import FixedMessage
 
 VariableFactorDist = Dict[str, Dict[Factor, AbstractMessage]]
 Projection = Dict[str, AbstractMessage]
+
+logger = logging.getLogger(
+    __name__
+)
 
 
 class MeanField(
@@ -434,19 +440,23 @@ class FactorApproximation(AbstractNode):
     ) -> Tuple["FactorApproximation", Status]:
         success, messages = Status() if status is None else status
 
-        factor_dist = (model_dist / self.cavity_dist)
-        if delta < 1:
-            log_norm = factor_dist.log_norm
-            factor_dist = (
-                    factor_dist ** delta * self.factor_dist ** (1 - delta))
-            factor_dist.log_norm = (
-                    delta * log_norm + (1 - delta) * self.factor_dist.log_norm)
+        try:
+            factor_dist = model_dist / self.cavity_dist
+            if delta < 1:
+                log_norm = factor_dist.log_norm
+                factor_dist = (
+                        factor_dist ** delta * self.factor_dist ** (1 - delta))
+                factor_dist.log_norm = (
+                        delta * log_norm + (1 - delta) * self.factor_dist.log_norm)
 
-        if not factor_dist.is_valid:
-            success = False
-            messages += (
-                f"model projection for {self} is invalid",)
-            factor_dist = factor_dist.update_invalid(self.factor_dist)
+            if not factor_dist.is_valid:
+                success = False
+                messages += (
+                    f"model projection for {self} is invalid",)
+                factor_dist = factor_dist.update_invalid(self.factor_dist)
+        except exc.MessageException as e:
+            logger.exception(e)
+            factor_dist = self.factor_dist
 
         new_approx = FactorApproximation(
             self.factor,
--- a/test_autofit/conftest.py
+++ b/test_autofit/conftest.py
@@ -37,25 +37,25 @@ def make_output_directory(
     return test_directory / "output"
 
 
-# @pytest.fixture(
-#     autouse=True,
-#     scope="session"
-# )
-# def remove_output(
-#         output_directory
-# ):
-#     yield
-#     for item in os.listdir(output_directory):
-#         if item != "non_linear":
-#             item_path = output_directory / item
-#             if item_path.is_dir():
-#                 shutil.rmtree(
-#                     item_path
-#                 )
-#             else:
-#                 os.remove(
-#                     item_path
-#                 )
+@pytest.fixture(
+    autouse=True,
+    scope="session"
+)
+def remove_output(
+        output_directory
+):
+    yield
+    for item in os.listdir(output_directory):
+        if item != "non_linear":
+            item_path = output_directory / item
+            if item_path.is_dir():
+                shutil.rmtree(
+                    item_path
+                )
+            else:
+                os.remove(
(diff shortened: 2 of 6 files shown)


autofit/graphical/mean_field.py
```diff
@@ -434,19 +440,23 @@ def project_mean_field(
...
-                    factor_dist ** delta * self.factor_dist ** (1 - delta))
-            factor_dist.log_norm = (
-                    delta * log_norm + (1 - delta) * self.factor_dist.log_norm)
-
-        if not factor_dist.is_valid:
-            success = False
-            messages += (
-                f"model projection for {self} is invalid",)
-            factor_dist = factor_dist.update_invalid(self.factor_dist)
+        try:
```
rhayes777, line 443: @matthewghgriffiths not sure if this is the best idea but it was necessary for easily testing other functionality using mock objects

test_autofit/conftest.py
```diff
@@ -37,25 +37,25 @@ def make_output_directory(
...
-                shutil.rmtree(
-                    item_path
-                )
-            else:
-                os.remove(
-                    item_path
-                )
+# @pytest.fixture(
+#     autouse=True,
+#     scope="session"
```
Jammy2211: ?

Jammy2211 (approved):
I am not sure if this is related to this specifric PR, but I just attempted to run the integration test and happened to Ctrl + C it really quick (by accident).

Anyway, this threw up the following error when I tried to resume the run:

```
  File "graphical/expectation_propagation.py", line 170, in <module>
    collection = factor_graph.optimise(laplace)
  File "/mnt/c/Users/Jammy/Code/PyAuto/PyAutoFit/autofit/graphical/declarative/abstract.py", line 135, in optimise
    updated_model = opt.run(
  File "/mnt/c/Users/Jammy/Code/PyAuto/PyAutoFit/autofit/graphical/expectation_propagation.py", line 365, in run
    model_approx, status = optimiser.optimise(factor, model_approx)
  File "/mnt/c/Users/Jammy/Code/PyAuto/PyAutoFit/autofit/non_linear/abstract_search.py", line 272, in optimise
    result = self.fit(
  File "/mnt/c/Users/Jammy/Code/PyAuto/PyAutoFit/autofit/non_linear/abstract_search.py", line 496, in fit
    self._fit(model=model, analysis=analysis, log_likelihood_cap=log_likelihood_cap)
  File "/mnt/c/Users/Jammy/Code/PyAuto/PyAutoFit/autofit/non_linear/nest/dynesty/abstract.py", line 120, in _fit
    sampler = self.paths.load_object(
  File "/mnt/c/Users/Jammy/Code/PyAuto/PyAutoFit/autofit/non_linear/paths/directory.py", line 86, in load_object
    return dill.load(
  File "/home/jammy/venvs/PyAuto/lib/python3.8/site-packages/dill/_dill.py", line 278, in load
    return Unpickler(file, ignore=ignore, **kwds).load()
  File "/home/jammy/venvs/PyAuto/lib/python3.8/site-packages/dill/_dill.py", line 481, in load
    obj = StockUnpickler.load(self)
EOFError: Ran out of input
```

The other thing to note is that we cannot currently give a path prefix for all results:

image

The `path_prefix` of the `DynestyStatic` search is not used (and it shouldn't be) but we probably want a `path_prefix` variable somewhere. Given we are probably going to assume all `AnalysisFactor`'s are in the same folder, we probably want to put it in `FactorGraphModel` or pass it to `factor_graph.optimise()`.

rhayes777:
Wait so you need to be able to customise the path prefix? Currently the path prefix is overridden to support this functionality

Jammy2211:
I need to be able to make it so all of the `AnalysisFactor0` `AnalysisFactor1` etc results are in the same folder, and ideally I wouldn't have to change the `output_path` to do this.

I don't think this should use the `path_prefix` of each individual `DynestyStatic` object, as then one could end up in a weird scenario where the `AnalysisFactor` outputs of the same fit end up in different folders.

I'm not sure where is the most appropriate place to put it.

rhayes777:
The other error is because a file was half-written I think

Jammy2211 (approved):

PR #407, opened 2021-11-11
feature/mp output

rhayes777:
- Log KL Diveregence and Evidence
- Plot KL Diveregence and Evidence
- Output a summary of the graph and contained models in graph.info

### DIFF
--- a/test_autofit/conftest.py
+++ b/test_autofit/conftest.py
@@ -37,25 +37,25 @@ def make_output_directory(
     return test_directory / "output"
 
 
-# @pytest.fixture(
-#     autouse=True,
-#     scope="session"
-# )
-# def remove_output(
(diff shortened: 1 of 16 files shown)


test_autofit/conftest.py
```diff
@@ -37,25 +37,25 @@ def make_output_directory(
     return test_directory / "output"
```
Jammy2211: Commented code

Jammy2211 (approved):
We should move all matplotlib plot calls and functions that do plots to a standalone plot package, like done at the end of this PR:

https://github.com/rhayes777/PyAutoFit/pull/410

This also helps a lot with the API for visualization when using the database (as a user can import and use Plotter objects without knowing the autofit source code API).

PR #412, opened 2021-11-17
Feature/search drawer

Jammy2211:
A Drawer non-linear search, which simply draws a fixed number of samples from the model uniformly from the
        priors.

        Therefore, it does not seek to determine model parameters which maximize the likelihood or map out the
        posterior of the overall parameter space.

        Whilst this is not the typical use case of a non-linear search, it has certain niche applications, for example:

        - Given a model one can determine how much variation there is in the log likelihood / log posterior values.
        By visualizing this as a histogram one can therefore quantify the behaviour of that
        model's `log_likelihood_function`.

        - If the `log_likelihood_function` of a model is stochastic (e.g. different values of likelihood may be
        computed for an identical model due to randomness in the likelihood evaluation) this search can quantify
        the behaviour of that stochasticity.

        - For advanced modeling tools, for example sensitivity mapping performed via the `Sensitivity` object,
        the `Drawer` search may be sufficient to perform the overall modeling task, without the need of performing
        an actual parameter space search.


autofit/non_linear/optimize/drawer/drawer.py
```diff
@@ -0,0 +1,178 @@
...
+            path_prefix=path_prefix,
+            unique_tag=unique_tag,
+            prior_passer=prior_passer,
+            initializer=initializer,
+            iterations_per_update=iterations_per_update,
+            session=session,
+            **kwargs
+        )
+
+        self.number_of_cores = 1
```
rhayes777, line 86: We could do it across multiple cores...

rhayes777 (approved):

PR #413, opened 2021-11-17
feature/sensitivity summary

rhayes777:


Jammy2211 (approved):
Could we have likelihood_base and likelihood_perturbed explicitly written out in the results.csv too, before likelihood difference?

I will be comparing sensitivity maps using different methods so its useful info!

PR #433, opened 2021-12-08
feature/fixes

rhayes777:


autofit/mapper/prior_model/abstract.py
```diff
@@ -609,7 +609,7 @@ def instance_from_prior_medians(self):
             A list of physical values
         """
         return self.instance_from_unit_vector(
-            unit_vector=[0.5] * len(self.prior_tuples)
```
rhayes777, line 612: This fixes a test: if priors are shared len(self.prior_tuples) > self.prior_count meaning that more values would be passed than was necessary. Don't think it had any real impact but got shown up by my newly added assertion

Jammy2211 (approved):

PR #436, opened 2021-12-10
Feature/factor plates

matthewghgriffiths:
I've refactored some of the Plate notation in Autofit to make it a little less brittle. In particular, Factors have to have plates associated with them (when necessary). All the current graphical tests past (as far as I can tell).

One oddity I noticed, there are 3 separate families of classes that are called 'Factor' but they don't all inherit from each other. 

With this refactor, I'm hopefully able to create a nice gradient API that's going to be based around the `autofit.mapper.operator.LinearOperator` and `autofit.graphical.transform.VariableTransform` classes (potentially with some refactoring of the latter class to more closely match the `autofit.mapper.operator.LinearOperator` class. 

I think the general implementation of the `autofit.graphical.transform.VariableTransform` should be called `autofit.graphical.transform.InstanceLinearOperator` and will allow linear transforms of 'model instances'. Where in this case the model instance is carrying gradient information.


autofit/graphical/factor_graphs/abstract.py
```diff
@@ -229,27 +282,27 @@ def _broadcast2d(
...
+    #     split into two properties to allow manual ordering 
+    #     of plate order
+    #     """
+    #     return self._plates or self.sorted_plates
+
+    # @cached_property
+    # def sorted_plates(self) -> Tuple[Plate]:
+    #     """
+    #     A tuple of the set of all plates in this graph, ordered by id
+    #     """
```
rhayes777, line 299: Remove commented code

autofit/graphical/factor_graphs/factor.py
```diff
@@ -630,27 +645,26 @@ def __call__(
...
         }
 
         if not (isinstance(res, tuple) or self.n_deterministic > 1):
             res = res,
 
-        log_val = (
-            0. if (shape == () or axis is None) else
-            aggregate(np.zeros(tuple(1 for _ in shape)), axis))
+        log_val = 0.
+        # log_val = (
```
rhayes777, line 662: Remove commented code

autofit/graphical/factor_graphs/abstract.py
```diff
@@ -160,6 +178,41 @@ def all_variables(self) -> Set[Variable]:
...
+
+        Returns
+        -------
+        The value reshaped to match the plates
+        """
+        shift = np.ndim(value) - len(plates)
+        if shift > 1 or shift < 0:
+            raise ValueError("dimensions of value incompatible with passed plates")
+        reduce_axes = tuple(
+            i + shift for i, p in enumerate(plates) if p and p not in self.plates)
```
rhayes777, line 206: There's no case in the tests where len(self.plates) > 0 meaning most of the functionality here is untested

autofit/graphical/factor_graphs/factor.py
```diff
@@ -273,12 +274,14 @@ def _function_shape(
...
-        var_shapes = {
-            k: np.shape(x) for k, x in kwargs.items()}
+        var_shapes = {k: np.shape(x) for k, x in kwargs.items()}
         return self._var_shape(**var_shapes)
 
     @lru_cache()
-    def _var_shape(self, **kwargs: Tuple[int, ...]) -> Tuple[int, ...]:
+    def _plate_sizes(
+        self, **kwargs: Tuple[int, ...]
+    ) -> Tuple[int, Dict[Union[None, str], Tuple[int, ...]]]:
```
rhayes777, line 283: That is a hell of a type annotation. Consider creating a class to comprise these variables

autofit/graphical/factor_graphs/factor.py
```diff
@@ -308,32 +311,45 @@ def _var_shape(self, **kwargs: Tuple[int, ...]) -> Tuple[int, ...]:
...
 
-        singleton dimensions are always assumed to match as in
-        standard array broadcasting
+            # Allow broadcasting of singleton dimensions 
+            elif ps == 1:
+                plate_sizes[p] = psize
+                return True 
+            elif psize == 1:
+                return True 
+            else:
```
rhayes777, line 328: No need for an else here as you return in every previous disjunct

autofit/graphical/factor_graphs/jacobians.py
```diff
@@ -52,16 +52,17 @@ def __init__(
             *,
             name="",
             vectorised=False,
-            is_scalar=False,
+            plates: Tuple[Plate, ...] = (),
             **kwargs: Variable
     ):
         self.vectorised = vectorised
-        self.is_scalar = is_scalar
+        self.is_scalar = bool(plates)
```
rhayes777, line 59: Surely this evaluates to True if len(plates) > 0 which I would have thought is the opposite of what you want?

autofit/graphical/factor_graphs/jacobians.py
```diff
@@ -327,26 +327,27 @@ def func_jacobian(
...
 
         if not (isinstance(vals, tuple) or self.n_deterministic > 1):
             vals = vals,
 
-        log_val = (
-            0. if (shape == () or axis is None) else
-            aggregate(np.zeros(tuple(1 for _ in shape)), axis))
+        log_val = 0.
+        # log_val = (
+        #     0. if (shape == () or axis is None) else
```
rhayes777, line 347: Commented code

autofit/graphical/factor_graphs/jacobians.py
```diff
@@ -327,26 +327,27 @@ def func_jacobian(
         kwargs = self.resolve_variable_dict(variable_dict)
         vals, *jacs = self._call_factor(
             kwargs, variables=variable_names)
-        shift, shape = self._function_shape(**kwargs)
-        plate_dim = dict(zip(self.plates, shape[shift:]))
-
+
```
rhayes777, line 330: Could you add more docs explaining the logic behind plates etc.?

autofit/mapper/variable.py
```diff
@@ -86,3 +88,55 @@ def ndim(self) -> int:
...
+
+
+def broadcast_plates(
+        value: np.ndarray, 
+        in_plates: Tuple[Plate, ...], 
+        out_plates: Tuple[Plate, ...], 
+        reducer: np.ufunc = np.sum
+    ) -> np.ndarray:
+    """
+    Extract the indices of a collection of plates then match
```
rhayes777, line 100: A more thorough explanation here would be helpful

autofit/mapper/variable.py
```diff
@@ -86,3 +88,55 @@ def ndim(self) -> int:
...
+
+
+    Returns
+    -------
+    The value reshaped to match the plates
+    """
+    n_in = len(in_plates)
+    n_out = len(out_plates)
+    shift = np.ndim(value) - n_in
+    if shift > 1 or shift < 0:
```
rhayes777, line 123: if not (0 <= shift <= 1) might be clearer

autofit/mapper/variable.py
```diff
@@ -86,3 +88,55 @@ def ndim(self) -> int:
...
+        raise ValueError("dimensions of value incompatible with passed plates")
+    
+    in_axes = list(range(shift, n_in + shift))
+    out_axes = []
+    k = n_out + shift
+    
+    for plate in in_plates:
+        try:
+            out_axes.append(out_plates.index(plate) + shift)
```
rhayes777, line 132: This enumerate/index/shift logic seems to be repeated. I wonder if there's a nice way to abstract it?

rhayes777 (approved):
It would be good to get some more detailed documentation on what plates are doing and also some explicit tests on the plate logic itself.

autofit/graphical/factor_graphs/abstract.py
```diff
@@ -111,7 +113,23 @@ def name_variable_dict(self) -> Dict[str, Variable]:
     @property
     @abstractmethod
     def deterministic_variables(self):
-        return self._deterministic_variables
+        ...
```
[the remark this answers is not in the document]
matthewghgriffiths: I think the main problem here is that we haven't defined a clean interface for 'Factors' in general - there are at least 3 different families of classes that are factor like. Is there a way that we can rationalise the interface for Factors? Perhaps as a protocol or a mixin?

autofit/graphical/factor_graphs/abstract.py
```diff
@@ -111,7 +113,23 @@ def name_variable_dict(self) -> Dict[str, Variable]:
...
+    def plates(self):
+        return self._plates
+
+    @cached_property
+    def sorted_plates(self) -> Tuple[Plate]:
+        """
+        A tuple of the set of all plates in this graph, ordered by id
+        """
+        return tuple(sorted(set(
+            cast(Plate, plate)
```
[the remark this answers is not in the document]
matthewghgriffiths: Yes, I copied this from some other bit of code, and I wasn't sure why we needed the cast, so I left it in. It's probably ok to delete?

autofit/graphical/factor_graphs/factor.py
matthewghgriffiths: That could be potentially worth it. I think we probably need to have a discussion about Plate notation in general, as I agree there's a lot of semi-duplicated functionality. `Tuple[int, Dict[Union[None, str], Int]]`

The key point of Plate functionality is to effectively keep track of NumPy dimensions.

autofit/graphical/factor_graphs/abstract.py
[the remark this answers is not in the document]
rhayes777: The declarative factors are distinct from factors in the graph so there's no need for them to share a protocol. It's fine for there to be reuse of names provided they are clearly separated by namespace

autofit/graphical/factor_graphs/jacobians.py
matthewghgriffiths: Yeah, that's a mistake

autofit/mapper/variable.py
matthewghgriffiths: yeah

autofit/graphical/factor_graphs/abstract.py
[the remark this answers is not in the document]
rhayes777: Usually it's to silence linter warnings about types so if there's no warning it doesn't need to be there

autofit/mapper/variable.py
matthewghgriffiths: I know what you mean, I think somewhere I use `None` as a sentinel value to indicated `shift = 1`. 

Again, this probably figures into refining the whole Plate notation API.

autofit/graphical/factor_graphs/abstract.py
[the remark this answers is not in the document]
matthewghgriffiths: There was a bit of Plate and Variable code that needed to get copied between each, so I'm not sure they're totally distinct?

Jammy2211 (approved):
The code looks fine other than what RIch commented.

I think we should define a toy-model / use case and build an example, which uses the core functionality of plates / gradients (and has scope for extension when we get there). 

I would be happy to lead the development of this once we've thought up a good example, as I don't think I'll understand this stuff otherwise!

matthewghgriffiths:
> 

Yeah that makes sense. 

Perhaps a good example would be the Hierarchical modelling? Currently, it doesn't use plates but is an example where it could be really useful.

matthewghgriffiths:
Though I think there might be an analytic/semi-analytic approach for the Hierarchical modelling

Jammy2211:
I have working examples for hierachical modeling:

https://github.com/Jammy2211/autofit_workspace/blob/master/scripts/howtofit/chapter_graphical_models/tutorial_4_hierachical.py

This essentially estimates the mean and sigma from which the `centre` parameters of data comprising 1D Gaussians are drawn. I have an analgous example for lensing.

Would the use of plates change the actual model, or just the API used in the example?

matthewghgriffiths:
Plates don't change the model (i.e. the posterior value should be the same), but they may change API and the quality of the approximations.

autofit/graphical/factor_graphs/abstract.py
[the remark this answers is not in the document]
rhayes777: Hmm maybe not in that case. That might be implementable with a mixin

autofit/mapper/operator.py
```diff
@@ -231,8 +250,96 @@ def __len__(self):
...
+        )
+
+    def __sub__(self, other: "MatrixOperator") -> "MatrixOperator":
+        return type(self).from_dense(
+            self.to_dense() - other.to_dense(), self.shape, self.ldim
+        )
+
+    def update(self, *args):
+        M = self.to_dense()
+        s = (Ellipsis,) + (None,) * self.ldim
```
rhayes777, line 327: What on earth is going on here?

rhayes777 (approved):
Definitely needs a lot more docs

autofit/graphical/utils.py
```diff
@@ -261,23 +253,16 @@ def invpsilog(c: np.ndarray) -> np.ndarray:
...
+            t.get_window_extent(renderer=r).transformed(ax.transData.inverted())
             for t in artists
         ]
-        min_extent = np.min(
-            [e.min for e in extents], axis=0
-        )
-        max_extent = np.max(
-            [e.max for e in extents], axis=0
-        )
+        min_extent = np.min([e.min for e in extents], axis=0)
```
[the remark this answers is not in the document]
matthewghgriffiths: No, because `e.min` is an vector/array. so `[e.min for e in extents]` is effectively a nx2 matrix that we want to find the min along both rows.

autofit/graphical/laplace/newton.py
```diff
@@ -0,0 +1,262 @@
...
+        fk_m1,
+        args=args,
+        search_direction=search_direction,
+        calc_line_search=calc_line_search,
+        line_search_kws=line_search_kws,
+        search_direction_kws=search_direction_kws,
+        f_count=f_count,
+        g_count=g_count,
+    )
+    Bk_next = quasi_newton_update(
```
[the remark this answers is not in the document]
matthewghgriffiths: The uppercase variables represent matrices, it's more confusing from a mathematical notation perspective to have matrices represented by lower case variables unfortunately.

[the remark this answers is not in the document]
rhayes777: Yeah I see that. I guess we should do whatever is best in terms of interpretting the code. From a Python perspective that would be strictly adhering to Python conventions but I can see how useful it could be to use cases in a more mathematical function.

autofit/graphical/utils.py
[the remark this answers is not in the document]
rhayes777: Ah I see

matthewghgriffiths:
Pull request superseeded by Jacobians pull request

PR #441, opened 2021-12-15
feature/ep result

rhayes777:


autofit/graphical/declarative/result.py
```diff
@@ -20,7 +19,7 @@ def __init__(
         ----------
         ep_history
             A history of the optimisation
-        declarative_factor
+        declarative_factor: AbstractDeclarativeFactor
```
Jammy2211, line 22: Type in _iinit_ and not docstring?

Jammy2211 (approved):

autofit/graphical/declarative/result.py
rhayes777: it's because it creates an import recursion

PR #446, opened 2021-12-22
feature/log uniform limits

rhayes777:


test_autofit/mapper/prior/test_limits.py
```diff
@@ -64,9 +64,14 @@ def test_vector_ignore_limits(model):
     )
 
 
-def test_log_uniform():
-    prior = af.LogUniformPrior()
-
+@pytest.mark.parametrize(
```
Jammy2211, line 67: You sure GaussianPriors which have optional upper and lower limits are being handled correctly too?

Jammy2211 (approved):

test_autofit/mapper/prior/test_limits.py
rhayes777: The default I used for testing is Gaussian. I guess I could make the tests slightly more concise and explicit by rolling GaussianPrior into this test

Jammy2211 (approved):

PR #454, opened 2022-01-08
Feature/pickle info

Jammy2211:
Provides info on a failed pickle file when scraping directories for database.


autofit/database/aggregator/scrape.py
```diff
@@ -273,10 +273,13 @@ def _add_pickles(
...
+        try:
+            with open(
+                    pickle_path / filename,
+                    "r+b"
+            ) as f:
+                fit[
+                    filename.split(".")[0]
+                ] = pickle.load(f)
+        except pickle.UnpicklingError:
+            raise pickle.UnpicklingError(f"Failed to unpickle: {pickle_path} {filename}")
```
rhayes777, line 285: Might be worth adding "from e" to that the original unpickling error message is propagated too

rhayes777 (approved):

Issue #450, opened 2021-12-26
update model / samples based on parameter names


Jammy2211:
**Use case:**

There are a number of circumstances where I want to produce a set of `Samples` with a `model` that removes certain parameters. For example, below I am writing a latex string for parameter results for a paper:

```
samples_result = list(agg_result_name.values("samples"))[0]
latex = af.text.Samples.latex(samples=samples_result)
```

This string contains every parameter in the model, whereas I only want to report 3 parameters in the paper and therefore wish to remove all other parameters.

Another use case is visualization, whereby I may wish to make plots of a subset of parameters.

**Solution:**

In both the `Model` and `Samples` classes we need methods that update both, removing parameters based on their names (which comes from the model path), for example:

`model = model.keep_parameters(parameter_names=["lens_bulge_intensity", "lens_mass_centre_0"]`
`samples = samples.keep_parameters(parameter_names=["lens_bulge_intensity", "lens_mass_centre_0"]`

Here, we are left with a model that has just two parameters remaining (`lens_bulge_intensity` and `lens_mass_centre_0`).

We can also remove parameters from a model:

`model = model.remove_parameters(parameter_names=["lens_disk"]`
`samples = samples.remove_parameters(parameter_names=["lens_disk"]`

Here, all parameters which have the string `lens_disk` are removed from the model and samples.

The strings input should use `in`, such that you dont have to specify the full parameter name to remove or keep it.

When we call these methods on `Samples` it should update the `model` internal to that `Samples` object by calling the corresponding method.

**Dimension Tracking**

To use in-built visualization libraries of certain searches, we store their results in their native format. This will be tricky to maniuplate with these methods (whereas for our owb `Samples` object removing parameters is trivial).

The simplest solution to this is dimension tracking, which gives us the information we need to use the visualization libraries in a way that behaves as expected. For example, if we had a `Samples` object whose model has 10 parameters and we wanted to make a plot of the `Dynesty` results with just 3 parameters corresponding to dimensions 2, 4 6 we would call the visualization tool as follows:

```
dyplot.cornerpoints(
    results=self.samples.results,
    labels=self.model.parameter_labels_with_superscripts_latex,
    ndim=[2, 4, 6],
    **kwargs
)
```

The array `self.samples.results` is the results in the native `Dynesty` format that are difficult for us to manipulate. 

[...]

PR #457, opened 2022-01-13
feature/with tuples

rhayes777:
Some improvements but I don't think it resolves the problem raised in #450

### DIFF
--- a/autofit/mapper/prior_model/abstract.py
+++ b/autofit/mapper/prior_model/abstract.py
@@ -26,6 +26,7 @@ from autofit.mapper.prior_model.recursion import DynamicRecursionCache
 from autofit.mapper.prior_model.util import PriorModelNameValue
 from autofit.text import formatter as frm
 from autofit.text.formatter import TextFormatter
+from autofit.tools.util import split_paths
 
 logger = logging.getLogger(
     __name__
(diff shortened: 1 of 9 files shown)


Jammy2211:
```
samples = samples.with_paths(
    ["galaxies.lens..mass.einstein_radius", "galaxies.source.bulge.sersic_index"]
)
```

Does the PR aim to support an API like this? (I thought I saw code for this).

I cant get it work using the above syntax. Is this syntax slightly wrong? And if so can we make it support this as I think this is the most concise syntax and is most in line with model composition.

autofit/mapper/prior_model/abstract.py
```diff
@@ -248,7 +251,7 @@ def _with_paths(
 
     def with_paths(
             self,
-            paths: List[Tuple[str]]
+            paths: List[Tuple[str, ...]]
```
Jammy2211, line 256: I am getting lint errors when there is more than 1 string in my tuple specifying paths.

autofit/mapper/prior_model/abstract.py
```diff
@@ -315,7 +321,7 @@ def _without_paths(
 
     def without_paths(
             self,
-            paths: List[Tuple[str]]
+            paths: List[Tuple[str, ...]]
```
Jammy2211, line 327: Same here.

autofit/mapper/prior_model/abstract.py
rhayes777: Yeah the ... should fix that

Jammy2211 (approved):

PR #465, opened 2022-01-23
change abc import to remove warning

Jammy2211:
The old import syntax was raising a depreciation warning when we ran unit tests.


rhayes777 (approved):
What was the warning? Deprecation?

Jammy2211:
> What was the warning? Deprecation?

Yep.

PR #470, opened 2022-01-31
Feature/sqlite optional

Jammy2211:
Well this is ugly.

### DIFF
--- a/autofit/database/model/prior.py
+++ b/autofit/database/model/prior.py
@@ -1,7 +1,5 @@
 from typing import Union
 
-from sqlalchemy import Column, Integer, ForeignKey
-
 from autofit.mapper.prior import abstract
 from autofit.mapper.prior_model import prior_model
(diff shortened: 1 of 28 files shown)


autofit/database/model/prior.py
```diff
@@ -57,15 +63,23 @@ class PriorModel(Object):
...
-    )
+        id = Column(
+            Integer,
+            ForeignKey(
+                "object.id"
+            ),
+            primary_key=True,
+        )
+
+    except ModuleNotFoundError:
```
rhayes777, line 80: alias sqlalchemy and then conditionally import that?

Jammy2211: The problem is because these are class attributes methods like:

```
id = Column(
            Integer,
            ForeignKey(
                "object.id"
            ),
            primary_key=True,
        )
```

Are called when autofit is imported. So they need to be in try / excepts. I think.

PR #472, opened 2022-02-01
Feature/jacobians

matthewghgriffiths:
I've set up a new Jacobian api which is mostly contained in `FactorJac` and backported to `Factor`. With this alterations I think we can get rid of `DeterminsticFactor`, `FactorJacobian`, `DeterminsticFactorJacobian`, and probably in the long run merge `Factor` and `FactorJac`.

I had some issues with test_autofit/graphical/hierarchical/test_embedded.py: test_full_fit, but I wasn't sure how to fix it ...

It's a bit of a monster request, Black was a bit aggressive, sorry...


rhayes777 (commented):
Mid review but I've undone the function indent formatting as that diagrees with PyCharm. We need to stick to consistent formatting otherwise every PR ever is going to be filled with indent changes

rhayes777:
I've resorted to local imports for some of the Jacobian stuff which was effectively circular. Bit annoying that it means we can't have type annotations but it's not an uncommon problem.

autofit/graphical/utils.py
```diff
@@ -1,20 +1,104 @@
...
+    return isinstance(v, Variable)
+
+
+def is_iterable(arg):
+    return isinstance(arg, collections.Iterable) and not isinstance(
+        arg, six.string_types
+    )
+
+
+def nested_filter(func, *args):
```
rhayes777, line 25: Probably worth adding docs

autofit/graphical/utils.py
```diff
@@ -44,45 +128,44 @@ def __init__(self, dict_: Dict[Variable, Tuple[int, ...]]):
...
-            arr[(ind,) * ndim] for ind in self.inds]
+        arrays = [arr[(ind,) * ndim] for ind in self.inds]
         arr_shapes = [arr.shape[ndim:] for arr in arrays]
-        return {
-            k: arr.reshape(shape * ndim + arr_shape)
-            if shape or arr_shape else arr.item()
-            for (k, shape), arr_shape, arr in
-            zip(self.items(), arr_shapes, arrays)}
+        return VariableData(
+            {
```
rhayes777, line 155: Given that VariableData only takes one argument I'd put the braces on the same line as the brackets

autofit/mapper/variable.py
```diff
@@ -86,3 +80,438 @@ def ndim(self) -> int:
...
+        newcls = super().__new__(cls, clsname, bases, attrs)
+        Variable.__init__(newcls, clsname)
+        return newcls
+
+
+class FactorValue(np.ndarray, metaclass=VariableMetaClass):
+    def __new__(cls, input_array, deterministic_values=None):
+        obj = np.asarray(input_array).view(cls)
+
+        if deterministic_values is None:
```
rhayes777, line 111: use or

autofit/mapper/variable.py
```diff
@@ -86,3 +80,438 @@ def ndim(self) -> int:
...
+def rtruediv(x, y):
+    return y / x
+
+
+class VariableData(Dict[Variable, np.ndarray]):
+    var_norm = _unary_op(np.linalg.norm)
+    var_det = _unary_op(np.linalg.det)
+    var_max = _unary_op(np.max)
+    var_min = _unary_op(np.min)
+    var_sum = _unary_op(np.sum)
```
rhayes777, line 277: neat

test_autofit/graphical/functionality/test_jacobians.py
```diff
@@ -0,0 +1,153 @@
+from itertools import combinations
+
+import pytest
+
+import numpy as np
+
+try:
+    import jax
```
rhayes777, line 8: We should install jax as part of the build and always execute these tests

test_autofit/graphical/functionality/test_jacobians.py
```diff
@@ -0,0 +1,153 @@
...
+        ),
+    ]
+
+    values = {x_: x, a_: a, b_: b, c_: c}
+    outputs = [factor.func_jacobian(values) for factor in factors]
+
+    tol = pytest.approx(0, abs=1e-4)
+    pairs = combinations(outputs, 2)
+    g0 = FactorValue(1.0, {z_: np.ones((5, 1))})
+    for (val1, jac1), (val2, jac2) in pairs:
```
rhayes777, line 82: Rather than a loop use [pytest.mark.parametrize](https://docs.pytest.org/en/6.2.x/parametrize.html)

test_autofit/graphical/hierarchical/test_embedded.py
```diff
@@ -133,42 +89,27 @@ def test_full_fit(centre_model, data, centres):
...
-            abs=0.1
-        )
+    optimiser = g.LaplaceOptimiser()
+
+    collection = graph.optimise(optimiser, max_steps=10).model
+
+    # TODO I don't know what's going on here?
+    # pred_centre = (
+    #     collection.HierarchicalFactor0.distribution_model.instance_from_prior_medians().mean
+    # )
```
rhayes777, line 108: It optimises a graph and makes an assertion about the expected output. It looks as though the values are coming out wrong. I think we saw similar arbitrary values before due to a windowing issue. This should pass

test_autofit/graphical/hierarchical/test_hierarchical.py
```diff
@@ -2,127 +2,309 @@
...
+    centres = centre_dist.rvs(n)
+    widths = np.exp(logt_dist.rvs(n)) ** -0.5
+    n_samples = np.random.poisson(n_samples_avg, size=n)
+
+    sample_dists = [stats.norm(*p) for p in zip(centres, widths)]
+    data = [dist.rvs(n) for dist, n in zip(sample_dists, n_samples)]
+    return data
+
+
+def test_full(data):
```
rhayes777, line 200: This test has no assertions

test_autofit/graphical/hierarchical/test_hierarchical.py
```diff
@@ -2,127 +2,309 @@
...
+    )
+
+
+def test_full_hierachical(data):
+    samples = {Variable(f"samples_{i}"): sample for i, sample in enumerate(data)}
+    x_i_ = [Variable(f"x_{i}") for i in range(n)]
+    logt_i_ = [Variable(f"logt_{i}") for i in range(n)]
+
+    mu_x_ = Variable("mu_x")
+    logt_x_ = Variable("logt_x")
```
rhayes777, line 244: Consider using fixtures for variables that are defined in mulitple tests

rhayes777 (approved):

matthewghgriffiths:
> I've resorted to local imports for some of the Jacobian stuff which was effectively circular. Bit annoying that it means we can't have type annotations but it's not an uncommon problem.

Why not have end of file imports instead?

rhayes777:
https://www.python.org/dev/peps/pep-0008/#:~:text=Imports%20are%20always%20put%20at%20the%20top%20of%20the%20file%2C%20just%20after%20any%20module%20comments%20and%20docstrings%2C%20and%20before%20module%20globals%20and%20constants.

To be fair my solution doesn't obey that either. I guess we could use end of file imports. I think it was applying autoformatting to undo the black indents that move the imports 🙃

matthewghgriffiths:
> Like, really circular imports indicate the code is tightly bound because one module should know about another and not vice-versa, but it can be nice in terms of API to have circular dependencies

What about using `TYPE_CHECKING` - https://www.python.org/dev/peps/pep-0563/#runtime-annotation-resolution-and-type-checking

rhayes777:
Interesting

rhayes777:
Probably a good idea to merge this and put any further changes on a new PR

rhayes777:
Avoid using git reset --hard to undo commits that have already been pushed as this changes the history for everyone. If you need to undo a specific commit then use git revert

matthewghgriffiths:
> 

Did I do this? I don't remember using that command!

rhayes777:
Not totally sure it's just that I had a 4 commit divergence between the two branches and I don't think I left any unpushed commits.

matthewghgriffiths:
Ah ok. I'm sure it was probably my mistake, still a bit of a novice git.

PR #479, opened 2022-02-18
Feature/stochastic

matthewghgriffiths:
I think this branch is ready to be merged into master. 

I've now started using these features to do actual model fitting on a large parameter model. Currently have a few issues with stability of optimisation, but this core functionality will be unaffected by that. 

Key additions to the API are the ability to index out of and into `EPMeanField`, `MeanField` and `AbstractMessage` values in place. This enables all the stochastic optimisation features used in `StochasticEPOptimiser`. Functionality is tested in the graphical/stochastic/test_regression.py test.


matthewghgriffiths:
This test fails, but I'm not sure why

https://github.com/rhayes777/PyAutoFit/blob/bc24b90b0bed716eeddbba56eac5aa0cd11f10ea/test_autofit/graphical/gaussian/test_optimizer.py#L54

autofit/graphical/expectation_propagation/ep_mean_field.py
```diff
@@ -65,6 +65,83 @@ def __init__(
...
+    def parameters(self):
+        return {
+            'factor_graph': self._factor_graph, 
+            'factor_mean_field': self._factor_mean_field
+        }
+
+    def copy(self):
+        return type(self)(**{k: val.copy() for k, val in self.parameters.items()})
+
+    def subset(self, plates_index):
```
rhayes777, line 78: Could you add some docs to this file?

autofit/graphical/expectation_propagation/ep_mean_field.py
```diff
@@ -223,3 +326,165 @@ def __repr__(self) -> str:
...
+        new_dist: MeanField,
+        factor_approx: FactorApproximation,
+        delta: float = 1.0,
+        status: Status = Status(),
+    ) -> Tuple["EPMeanField", Status]:
+
+        factor_mean_field = self.factor_mean_field
+        
+        # We're fitting the full factor_dist, not the subset factor_dist
+        rescale1 = {
```
rhayes777, line 425: Why is this rescaling necessary?

autofit/graphical/expectation_propagation/ep_mean_field.py
```diff
@@ -223,3 +326,165 @@ def __repr__(self) -> str:
...
+            factor_graph=self._factor_graph,
+            factor_mean_field=factor_mean_field,
+            factor_rescale=self._factor_rescale,
+            factor_subset_factor=self._factor_subset_factor, 
+            ep_mean_field=self._ep_mean_field,
+            plates_index=self._plates_index,
+        )
+        return new_approx, status
+
+    def factor_approximation(self, factor: Factor) -> FactorApproximation:
```
rhayes777, line 450: Am I confused or is something similar computed elsewhere?

autofit/graphical/expectation_propagation/optimiser.py
```diff
@@ -55,6 +59,28 @@ def optimise(
...
+            new_approx = model_approx if self.inplace else model_approx.copy()
+            new_approx.update_factor_mean_field(factor, factor.calc_exact_update(factor_approx.cavity_dist))
+        elif factor._calc_exact_projection:
+            factor_approx = model_approx.factor_approximation(factor)
+            new_model_dist = factor.calc_exact_projection(factor_approx.cavity_dist) 
+            new_approx, status = self.update_model_approx(
+                new_model_dist, factor_approx, model_approx, status
+            )
+        else:
+            raise NotImplementedError
```
rhayes777, line 76: Add a message indicating why the error was raised

autofit/graphical/expectation_propagation/optimiser.py
```diff
@@ -155,12 +182,51 @@ def __init__(
...
+        factor_graph = model_approx.factor_graph 
+        factor_order = factor_order or factor_graph.factors
+
+        factor_optimisers = factor_optimisers or {}
+        factor_mean_field = model_approx.factor_mean_field
+        for factor in factor_graph.factors:
+            if factor not in factor_optimisers:
+                factor_dist = factor_mean_field[factor]
+                if factor.has_exact_projection(factor_dist):
+                    factor_optimisers[factor] = ExactFactorFit(**(exact_fit_kws or {}))
```
rhayes777, line 213: It would probably make more sense to optionally pass in an ExactFactorFit rather than optionally pass in kwargs

autofit/graphical/expectation_propagation/optimiser.py
```diff
@@ -278,3 +354,118 @@ def _output_results(self, model_approx: EPMeanField):
...
+            self, 
+            factor, 
+            subset_approx, 
+            optimiser=None
+    ):
+        factor_logger = self.factor_loggers.get(factor.name, logging.getLogger(factor.name))
+        factor_logger.debug("Optimising...")
+
+        optimiser = optimiser or self.factor_optimisers[factor]
+        subset_factor = subset_approx._factor_subset_factor[factor]
```
rhayes777, line 371: So the subset of factors remains the same each factor step?

autofit/graphical/factor_graphs/factor.py
```diff
@@ -198,13 +202,80 @@ def __init__(
...
+
+        return Factor(subset_factor, *self.args, **kws)
+
+    def _subset_jacobian(self, subset_factor, index):
+        jac_kws = {"vjp": self._vjp}
+        if self._vjp:
+            factor_vjp = None
+            if self._factor_vjp != self._jax_factor_vjp:
+                try:
+                    factor_vjp = self._factor_vjp[index]
```
rhayes777, line 244: Create a local function which executes this try except pattern on a generic object to avoid repetition

autofit/graphical/mean_field.py
```diff
@@ -260,6 +323,62 @@ def from_dist(
...
+                # e.g. factor_dist.check_valid().sum() / factor_dist.check_valid().size
+                valid = factor_dist.check_valid()
+                if valid.any():
+                    updated = True
+                    n_valid = valid.sum()
+                    n_total = valid.size
+                    logger.debug(
+                        "meanfield with variables: %r ,"
+                        "partially updated %d parameters "
+                        "out of %d total, %.0%%",
```
rhayes777, line 363: You can do formatting with fstrings

autofit/graphical/utils.py
```diff
@@ -362,3 +374,53 @@ def rescale_to_artists(artists, ax=None):
...
+def aggregate(array: np.ndarray, axis: Axis = None, **kwargs) -> np.ndarray:
+    """
+    aggregates the values of array
+
+    if axis is False then aggregate returns the unmodified array
+
+    otherwise aggrate returns np.sum(array, axis=axis, **kwargs)
+    """
+    if axis is False:
+        return array
```
rhayes777, line 410: No need for else when if returns

autofit/mapper/operator.py
```diff
@@ -609,7 +619,19 @@ def __init__(self, scale, inv_scale=None):
...
-        self._ldim = len(self.scale.shape)
+        self._ldim = self.scale.ndim
+
+    def __getitem__(self, index):
+        if index[:self.ldim] == index[self.ldim:]:
+            return DiagonalMatrix(
+                self.scale[index[:self.ldim]]
+            )
+        else:
+            raise NotImplementedError()
```
rhayes777, line 630: It's work adding a message explaining this.

Also this should probably be a ValueError

autofit/messages/abstract.py
```diff
@@ -456,14 +505,22 @@ def as_factor(self, variable: "Variable", name: Optional[str] = None):
...
             factor_jacobian=self.factor_jacobian,
             plates=variable.plates,
             arg_names=["x"],
         )
 
+    def calc_exact_update(self, x: "AbstractMessage") -> "AbstractMessage":
+        return self, 
+
+    def has_exact_projection(self, x: "AbstractMessage") -> bool:
+        if type(self) == type(x):
```
rhayes777, line 520: return if type(self) == type(x)

test_autofit/graphical/stochastic/test_regression.py
```diff
@@ -0,0 +1,105 @@
...
+
+x = 5 * np.random.randn(n_obs, n_features)
+y = x.dot(a) + b + np.random.randn(n_obs, n_dims)
+
+obs = graph.Plate(name='obs')
+features = graph.Plate(name='features')
+dims = graph.Plate(name='dims')
+
+x_ = graph.Variable('x', obs, features)
+a_ = graph.Variable('a', features, dims)
```
rhayes777, line 28: These would make lovely fixtures

rhayes777 (approved):

rhayes777:
test_laplace_jac is taking way too long to execute

rhayes777:
I've fixed the failing test. AbstractMessage ids needed to be passed when copying as Priors are identified by their associated id. Copying resulted in a mismatch between ids

autofit/graphical/mean_field.py
```diff
@@ -441,52 +487,57 @@ def project_mean_field(
...
             self.cavity_dist,
             factor_dist=factor_dist,
             model_dist=model_dist,
         )
-        return new_approx, Status(
-            success=success, messages=messages, flag=flag, updated=updated
-        )
+        return new_approx, status
+
+        # success, messages, _, flag = Status() if status is None else status
```
Jammy2211: ?

autofit/graphical/mean_field.py
```diff
@@ -132,25 +132,25 @@ def subset(self, variables=None, plates_index=None, rescale=False):
...
-            def make_indexes(v, plates_index, plate_sizes):
-                if any(p in plates_index for p in v.plates):
-                    return np.ix_(
-                        *(
-                            make_index_seq(p, plates_index, plate_sizes)
-                            for p in v.plates
-                        )
-                    )
-                return ()
+            # def make_index_seq(p, plates_index, plate_sizes):
```
Jammy2211: delete?

autofit/graphical/expectation_propagation/optimiser.py
```diff
@@ -43,10 +44,18 @@ def update_model_approx(
             status: Optional[Status] = Status(),
     ) -> Tuple[EPMeanField, Status]:
         delta = self.deltas[factor_approx.factor]
-        projection, status = factor_approx.project(
-            new_model_dist, delta=delta, status=status
+        # mean_field, status = factor_approx.project(
+        #     new_model_dist, delta=delta, status=status
```
Jammy2211: delete?

Jammy2211 (approved):

autofit/graphical/expectation_propagation/ep_mean_field.py
matthewghgriffiths: *waves hands at maths*

The rescaling is necessary as we're effectively factorising the subfactor out of the factor, so we have to account for that when calculating the cavity and factor distributions. 

I'm not 100% confident that the rescaling has to be done per variable as opposed to per factor but the tests kinda work.

autofit/graphical/expectation_propagation/optimiser.py
matthewghgriffiths: I'm not sure what you mean by this question. The subset of factors changes every batch.

autofit/graphical/mean_field.py
matthewghgriffiths: When logging it's better to use c string formatting as then the interpolation only happens when the logging occurs.

https://docs.python.org/3/howto/logging.html#logging-variable-data

rhayes777: That link is broken for some reason. 503 so I think the server is down. *shakes fist at Russia*

That seems to make sense for debug level logging in a part of the code that gets called a lot. I suspect for the most part string formatting for logs has little impact on execution time so I'd go for clarity over speed.

autofit/graphical/expectation_propagation/ep_mean_field.py
matthewghgriffiths: Yes, but not for EPMeanFieldSubset objects, so the calculation is subtly different.

autofit/graphical/expectation_propagation/optimiser.py
matthewghgriffiths: That potentially makes sense, I'm finding that I'm having to write custom optimisation loops, so I'm not using this class very much.

autofit/graphical/mean_field.py
matthewghgriffiths: Is it really that much less readable?

rhayes777: To my mind it is because you have to cross-match to see where variables go

test_autofit/graphical/stochastic/test_regression.py
matthewghgriffiths: Fixtures are harder for me to debug!

rhayes777:
If it doesn't break anything it's probably sensible to merge this now so new changes can be seen in a new PR

PR #498, opened 2022-03-30
feature/domain

### LINKED ISSUE #495: Out of bound parameters when using Gaussian priors during hierarchical model fitting
I am fitting a hierarchical model using parametric density profiles to several different profiles measured directly from simulations. The profile is fully specified using 5 parameters. My priors for three of said parameters are:
image

Once I start the fitting via factor_graph.optimise(), which uses EP, the procedure crashes and prints the following error:
image

@Jammy2211 and I investigated the cause, and it seems like the bounds on the Gaussian priors are not being enforced. For example, this is a set of parameters drawn by the optimiser: 
image
image

In same cases, it leads to complex densities being predicted:  
image

Performing the fit using uniform priors instead of Gaussian works better, but the optimiser does not converge to an answer. Fitting each density profile individually (no EP, graphical or hierarchical modelling) works fine, regardless of whether Gaussian or Uniform priors are used.

rhayes777:
Ensure that limits are copied through other operations on messages.

Should fix #495

### DIFF
--- a/autofit/messages/abstract.py
+++ b/autofit/messages/abstract.py
@@ -421,7 +421,7 @@ class AbstractMessage(Prior, ABC):
 
     @classmethod
     def from_mode(
-            cls, mode: np.ndarray, covariance: np.ndarray, id_
+            cls, mode: np.ndarray, covariance: np.ndarray, **kwargs
     ) -> "AbstractMessage":
         pass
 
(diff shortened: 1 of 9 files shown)


autofit/messages/abstract.py
```diff
@@ -268,7 +270,7 @@ def _broadcast_natural_parameters(self, x):
             )
 
     def factor(self, x):
-       # self.assert_within_limits(x)
+        # self.assert_within_limits(x)
```
Jammy2211, line 300: Still cant tell if we should be setting up the bounds ofr the factor optimization properly here or not...

Anyway, if the test works this is fine for now.

Jammy2211 (approved):

PR #512, opened 2022-04-11
Multiple hierarchical factor fits: fix for sorting of hierarchical_factor_set 

VictorForouhar:
Previously, trying  to fit a hierarchical model with more than one hierarchical factor would return the following error:
image
When fitting just one hierchical factor this error does not happen because sorted() does not need to compare between two factors. 

I added a way to correctly sort the hierarchical factor list by name so that it does not return an error. 
image


rhayes777 (approved):
It may make sense to implement this using `__lt__` and `__gt__` in the HierarchicalFactor class

PR #517, opened 2022-05-04
feature/composition

rhayes777:
Priors now contain a message rather than inheriting from one. This allows prior arithmetic to be used.

i.e.
```python
af.Add(left, right)
```

can be written
```python
left + right
```

Also adds a fix for variable names in relations


test_autofit/graphical/test_composition.py
```diff
@@ -26,8 +26,8 @@ def test_gaussian_prior():
     "cls",
     [
         af.UniformPrior,
-        af.LogUniformPrior,
-        af.LogGaussianPrior,
+        # af.LogUniformPrior,
```
Jammy2211, line 29: ?

test_autofit/graphical/test_composition.py
```diff
@@ -37,10 +37,10 @@ def test_other_priors(cls):
         prior,
         AbstractMessage
     )
-    assert isinstance(
-        prior.message,
-        AbstractMessage
-    )
+    # assert isinstance(
```
Jammy2211, line 40: ?

autofit/mapper/prior/prior.py
```diff
@@ -57,12 +57,13 @@ def _new_for_base_message(
             self,
             message
     ):
+        print(type(self))
```
Jammy2211: ?

autofit/messages/abstract.py
```diff
@@ -462,7 +461,7 @@ def log_normalisation(self, *elems: Union["AbstractMessage", float]) -> np.ndarr
     @staticmethod
     def _iter_dists(dists) -> Iterator[Union["AbstractMessage", float]]:
         for elem in dists:
-            if isinstance(elem, Prior):
+            if isinstance(elem, AbstractMessage):
```
Jammy2211: Both yield same thing>

Jammy2211 (approved):

autofit/messages/abstract.py
rhayes777: Priors are no longer messages

PR #532, opened 2022-05-12
feature/hierarchical median

rhayes777:
Implements function to recover the latest result for a factor from EPResult

```python
result = ep_result.latest_for(factor)
```

If the factor is a hierarchical factor then a HierarchicalResult is returned.

Hierarchical results comprise individual results produced by the latest optimisation of each associated
factor. One factor is created for each variable drawn from the distribution.

Models are created by taking the product of the posteriors of the final optimisations.
Instances are created from a combined samples set which includes all samples from the comprised results.


Jammy2211:
Does this work for HierarchicalFactors in the not EP case?

Jammy2211 (approved):

rhayes777:
No I don't think so

Issue #544, opened 2022-06-02
Log evidences of EP fit are zero


Jammy2211:
If I run a simple EP fit:

https://github.com/Jammy2211/autofit_workspace_test/blob/main/ep_examples/toy/model/ep.py

The Log Evidence values printed throughout the fit are being given as 0:

```
2022-06-02 16:49:11,208 - dataset_9 - INFO - Log Evidence = 0.0
2022-06-02 16:49:11,208 - dataset_9 - INFO - KL Divergence = 194184846802789.72
2022-06-02 16:49:12,018 - PriorFactor2 - INFO - Log Evidence = 0.0
2022-06-02 16:49:12,019 - PriorFactor2 - INFO - KL Divergence = 194184846802789.72
2022-06-02 16:49:12,221 - PriorFactor3 - INFO - Log Evidence = 0.0
2022-06-02 16:49:12,221 - PriorFactor3 - INFO - KL Divergence = 194184846802777.9
2022-06-02 16:49:13,288 - PriorFactor30 - INFO - Log Evidence = 0.0
2022-06-02 16:49:13,289 - PriorFactor30 - INFO - KL Divergence = 194184846802775.25
```

The `graph.png` is also reflecting this behaviour:

![image](https://user-images.githubusercontent.com/23455639/171670692-a9b1ab91-bb45-418e-bbf3-db4a6a8e35aa.png)

Jammy2211:
Shouldn't the KL Divergence also be decreasing?

PR #547, opened 2022-06-06
feature/zero evidence

rhayes777:
Ensure log evidence is non-zero by including the self message in the computation for #544


rhayes777:
This breaks message tests against log_normalisation so I assume isn't a valid solution. Any thoughts @matthewghgriffiths ?

Jammy2211 (approved):

PR #551, opened 2022-06-20
Feature/result info str

Jammy2211:
The `model.info` file string is now also an attribute of `result.info`.

### DIFF
--- a/autofit/non_linear/paths/abstract.py
+++ b/autofit/non_linear/paths/abstract.py
@@ -414,13 +414,19 @@ class AbstractPaths(ABC):
             json.dump(config_dict, f, indent=4)
 
     def save_summary(self, samples, log_likelihood_function_time):
-        text_util.results_to_file(
+
+        result_info = text_util.result_info_from(
             samples=samples,
             filename=path.join(
                 self.output_path,
                 "model.results"
             )
         )
+
+        filename = path.join(self.output_path, "model.results")
+
+        with open_(filename, "w") as f:
+            f.write(result_info)
 
         text_util.search_summary_to_file(
             samples=samples,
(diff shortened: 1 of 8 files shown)


autofit/non_linear/paths/abstract.py
```diff
@@ -414,14 +414,16 @@ def _save_search(self, config_dict):
...
+
+        result_info = text_util.result_info_from(
             samples=samples,
-            filename=path.join(
-                self.output_path,
-                "model.results"
-            )
         )
 
+        filename = path.join(self.output_path, "model.results")
```
rhayes777, line 422: We should start using pathlib rather than joining paths

docs/features/search_chaining.rst
```diff
@@ -56,19 +56,41 @@ this give a speed-up in log likelihood evaluation.
   :width: 600
   :alt: Alternative text
 
+We first compose the model, which only represents the left hand `Gaussian`.
```
rhayes777, line 59: What's the advantage of RST over Markdown?

rhayes777 (approved):

docs/features/search_chaining.rst
Jammy2211: readthedocs doesn't support Markdown

rhayes777: Ah right

Jammy2211: Ok maybe it does, anyway we're way too far gone to swap now haha

autofit/non_linear/paths/abstract.py
Jammy2211: Yeah I need to learn pathlib.

docs/features/search_chaining.rst
Jammy2211: Have you seen our fancy new readthedocs theme btw?

https://pyautofit.readthedocs.io/en/latest/

PR #555, opened 2022-06-23
method added for making random instance

Jammy2211:


rhayes777 (approved):
Might it not make more sense to use the intrinsic limits of each individual prior?

Jammy2211:
I'm not sure, the problem is when we generate values from `GaussianPrior`'s and they go beyond their limits. 

We do not have a method to generate values within these limits for a Gaussian, and it changes the properties where it would probably make more sense that we implement a `GaussianTruncatedPrior`.

rhayes777:
Could we use a guassian PDF normalised between two limits to transform a number 0->1 to a number on a 'truncated' Gaussian?

Jammy2211:
Yes we could, but I wouldn't want to mix that in with the `GaussianPrior` because they are different distributions.

rhayes777:
We have a PDF associated with each message so it might be a method we could implement in general for priors

Jammy2211:
Yeah, its definitely where we need to go.

We need to think about how it links to prior passing as well though, as really we should be linking priors which generate points based on the limits in this way too. Hence why I avoided going to deep into this on this PR lol.

PR #564, opened 2022-07-11
added free parameters etc. to model info

### LINKED ISSUE #561: print(model.info) does not show number of parameters at top
Which is done when you read a `model.info` file, probably means that string is written separately.

rhayes777:
Resolves #561


rhayes777:
It's not HTML so I don't think so. We could just reduce the padding?

PR #566, opened 2022-07-11
feature/laplace dynamic delta

rhayes777:
Better unifies the optimize function between Laplace and other optimisers. 

This allows the delta computed from the number of shared factors for a given variable to be applied in the case of Laplace. It also illustrated the in the case of classical autofit optimisers a projection was applied to a factor then to the model; in the case of Laplace it only seems that projection is applied to the model. Classical optimisers now implement the same behaviour as Laplace.

### DIFF
--- a/autofit/graphical/expectation_propagation/factor_optimiser.py
+++ b/autofit/graphical/expectation_propagation/factor_optimiser.py
@@ -45,13 +45,12 @@ class AbstractFactorOptimiser(ABC):
                 for variable, message_count in variable_message_count.items()
             })
 
-        new_approx, status = model_approx.project_mean_field(
+        return model_approx.project_mean_field(
             new_model_dist,
             factor_approx,
             delta=delta,
             status=status,
         )
-        return new_approx, status
 
     @abstractmethod
     def optimise(


autofit/non_linear/abstract_search.py
```diff
@@ -289,25 +287,12 @@ def optimise(
...
-
-        if self.dynamic_delta:
-            delta = MeanField({
-                variable: self.delta * (min_value / message_count)
-                for variable, message_count in variable_message_count.items()
-            })
-        else:
-            delta = self.delta
-
-        projection, status = factor_approx.project(
```
rhayes777, line 303: This projection step does not occur anymore. @matthewghgriffiths is this correct?

autofit/graphical/expectation_propagation/factor_optimiser.py
```diff
@@ -16,11 +16,12 @@ class AbstractFactorOptimiser(ABC):
     """
     logger = logger.debug
 
-    def __init__(self, initial_values=None, deltas=None, inplace=False, delta=1):
+    def __init__(self, initial_values=None, deltas=None, inplace=False, delta=1, dynamic_delta=False):
```
rhayes777, line 19: Currently the default optimiser would need to be passed with dynamic_delta=True explicitly. I'm tempted to make the default dynamic_delta=True

Jammy2211: Once I test this feature set properly I suspect we'll end up making it True by default.

Jammy2211 (approved):

autofit/graphical/expectation_propagation/factor_optimiser.py
rhayes777: Yeah makes sense. I also need to look at incorporating it into the Laplace optimiser

rhayes777: Oh wait that's what this PR does lol

autofit/non_linear/abstract_search.py
```diff
@@ -289,25 +287,12 @@ def optimise(
...
-            new_model_dist,
-            delta=delta
+        model_approx, status = self.update_model_approx(
+            new_model_dist, factor_approx, model_approx, status
         )
-
         status.result = result
```
rhayes777, line 294: test

autofit/graphical/expectation_propagation/factor_optimiser.py
```diff
@@ -16,11 +16,12 @@ class AbstractFactorOptimiser(ABC):
...
         self.initial_values = initial_values or {}
         self.inplace = inplace
         self.delta = delta
         self.deltas = deltas or {}
+        self.dynamic_delta = dynamic_delta
 
     def update_model_approx(
```
matthewghgriffiths, line 26: Suggest deleting this method and merging functionality into EPOptimiser class

autofit/graphical/expectation_propagation/factor_optimiser.py
```diff
@@ -30,14 +31,26 @@ def update_model_approx(
...
+        return model_approx.project_mean_field(
             new_model_dist,
             factor_approx,
             delta=delta,
             status=status,
         )
-        return new_approx, status
 
     @abstractmethod
     def optimise(
```
matthewghgriffiths, line 56: This method should have signature,

```
def optimise(
    self, factorApprox: FactorApproximation
) -> Tuple[FactorApproximation, Status]:
```

matthewghgriffiths (commented):
Finishing review

PR #568, opened 2022-07-25
Feature/result str

Jammy2211:
Builds on this PR:

https://github.com/rhayes777/PyAutoFit/pull/564

1) Makes LINE_LENGTH a config file parameter, as users may need different values for optimal display.

2) Applies to more TextFormatters, so results also use this parameter.


autofit/text/samples_text.py
```diff
@@ -24,6 +26,8 @@ def summary(
     sigma
         The sigma within which the PDF is used to estimate errors (e.g. sigma = 1.0 uses 0.6826 of the PDF)."""
 
+    LINE_LENGTH = conf.instance["general"]["output"]["info_whitespace_length"]
```
rhayes777, line 29: Might be worth making a function rather than repeating this

test_autofit/text/test_text_util.py
```diff
@@ -37,7 +37,7 @@ def test__results_to_file(samples):
         samples=samples,
     )
 
-    assert  "Maximum Log Likelihood                                                                    1.00000000\n" in result_info
+    assert  "Maximum Log Likelihood                                                          1.00000000\n" in result_info
```
rhayes777, line 40: I wonder if it's worth making this test dynamic

rhayes777 (approved):

test_autofit/text/test_text_util.py
Jammy2211: Don't think so, the config is a tesrt specific one. Worry that dynamic might miss changes to source code.

PR #577, opened 2022-09-28
Feature/versions

Jammy2211:
Updating many versions to latest, which were hard coded for a long time.

Will verify all dependencies are ok on build server once on master.


autofit/non_linear/nest/dynesty/abstract.py
```diff
@@ -96,6 +96,9 @@ def resample_figure_of_merit(self):
              -np.inf is an invalid sample value for Dynesty, so we instead use a large negative number."""
             return -1.0e99
 
+        def history_save(self):
+            pass
```
rhayes777, line 100: If this is an abstract class maybe make this an abstractmethod?

rhayes777 (approved):

PR #581, opened 2022-10-03
fix test causing build to fail because a message cannot be pickled

rhayes777:


Jammy2211:
This actually stops actual Dynesty runs from working lol

rhayes777:
Hmm will try and solve the underlying issue

rhayes777:
Ah I see so you readd it after. I guess that's a solution

Jammy2211:
There was no underlying issue we literally removed dynesty's likelihood function lol

rhayes777:
I mean as in when using dill to serialise a factor

PR #586, opened 2022-10-10
feature/mp api

rhayes777:
Make the factor optimiser's optimise method take a factor approximation and return a new model distribution. This is a prerequisite for parallelisation at the factor optimisation level.


rhayes777:
Currently tests fail for stochastic and exact fit; it's not totally clear how to fix these @matthewghgriffiths ?

autofit/non_linear/abstract_search.py
```diff
@@ -257,9 +255,9 @@ def optimise(
                 f" AnalysisFactors, HierarchicalFactors and PriorFactors"
             )
 
-        factor_approx = model_approx.factor_approximation(
-            factor
-        )
+        # factor_approx = factor_approx.factor_approximation(
```
Jammy2211, line 258: ?

Jammy2211 (approved):

matthewghgriffiths (approved):

PR #587, opened 2022-10-10
Feature/docs2

Jammy2211:


docs/conf.py
```diff
@@ -21,6 +21,16 @@
 
 sys.path.insert(0, os.path.abspath("."))
 
+clone_path = str(here())
```
rhayes777, line 24: This would probably be easier using pathlib.Path

rhayes777 (approved):

PR #589, opened 2022-10-14
Feature/rstate fix

Jammy2211:
hotfix for missing rstate in dynestty


autofit/non_linear/nest/dynesty/abstract.py
```diff
@@ -179,6 +179,9 @@ def _fit(self, model: AbstractPriorModel, analysis, log_likelihood_cap=None):
                 config_dict_run = self.config_dict_run
                 config_dict_run.pop("maxcall")
 
+                if not hasattr(sampler, "rstate"):
```
rhayes777, line 182: Seems pretty funky to attach this function on the fly

rhayes777 (approved):

autofit/non_linear/nest/dynesty/abstract.py
Jammy2211: Its a weird hotfix, I dont totally follow why Dynesty is losing it and cant reproduce it locally but hopefully we'll catch it with the build server somehow.

PR #591, opened 2022-10-19
Feature/dynesty update

Jammy2211:
Update dynesty wrapper to be compatible with `dynesty==2.0.1`.

Dynesty 2.0.1 introduces in built check pointing:

Section "Checkpointing" at https://dynesty.readthedocs.io/en/stable/quickstart.html

This PR updates the autofit wrapper to use this checkpoint, whereas previously we would pickle the dynesty sampler and object ourselves. This should not change functionality much, but using in-built dynesty checkpointing may improve sampling (as I was already worried our method lost information).

@rhayes777 Dynesty also updated multiprocessing, Sections "Parallelization" and "Dynesty multiprocessing pool", meaning that the sneaky pool no longer works. Could you size up the new dynesty implementation and either update sneaky pool to work or use their new implementation (which appears to support passing large objects to each individual LH function).


autofit/non_linear/nest/dynesty/dynamic.py
```diff
@@ -80,16 +103,19 @@ def sampler_from(
...
-            prior_transform=prior_transform,
-            ndim=model.prior_count,
-            logl_args=[model, fitness_function],
-            ptform_args=[model],
-            queue_size=self.number_of_cores,
-            pool=pool,
-            **self.config_dict_search
-        )
+        try:
+            return DynamicNestedSampler.restore(fname=self.checkpoint_file, pool=pool)
```
rhayes777, line 107: What stops us from passing in sneaky pool here?

autofit/non_linear/nest/dynesty/static.py
```diff
@@ -73,30 +75,57 @@ def __init__(
...
-            ptform_args=[model],
-            live_points=live_points,
-            queue_size=self.number_of_cores,
-            pool=pool,
-            **self.config_dict_search
-        )
+        try:
+            return StaticSampler.restore(
+                fname=self.checkpoint_file,
+                pool=pool
```
rhayes777, line 114: And here?

rhayes777 (approved):

PR #594, opened 2022-10-28
feature/symlink

rhayes777:
Remove Symlink creation


autofit/non_linear/abstract_search.py
```diff
@@ -14,7 +14,12 @@
 
 from autofit import exc
```
Jammy2211, line 15: BLACK !? :O

rhayes777: Yeah they use it at my client so I got used to having it as a filewatcher which is convenient. Also I figured out how to make it format how I want

Jammy2211 (approved):

PR #601, opened 2022-11-21
feature/mpmp

rhayes777:
Includes commits from a currently open PR that should be merged.

- Removes delta from optimisers
- Implements the ApproxUpdater class to handle updating of the mean field inside the EPOptimiser rather than factor optimisers
- Implements a basic ParallelEPOptimiser that maps factor optimisation to separate processes

# ApproxUpdater

An ApproxUpdater is now passed to the EPOptimiser class to dictate how updates occur:

```python
opt = graph.EPOptimiser(model_approx.factor_graph, updater=SimpleUpdater(delta=1.0))
```

Options include:

- SimplerUpdater - set a float value for delta
- FactorUpdater - set a delta value on a factor by factor basis
- DynamicUpdater - dynamically computes delta so variables shared by more factors update more slowly

# ParallelEPOptimiser

This is the same as a regular EPOptmiser but the number of available cores can be passed in.

```python
opt = ParallelEPOptimiser(
    model_approx.factor_graph,
    n_cores=len(model_approx.factors) + 1,
)
```

The number of cores used by the pool is 1 less than the number passed in as it is assumed one core is used by the main process. A number of cores less than 3 will induce an error.

All factors are optimised simultaneously before the EPMeanField is updated sequentially. It is not guaranteed to be faster than non-parallel optimisation; speed improvements should be tessted and further optimisation may be required.


autofit/mapper/prior/abstract.py
```diff
@@ -37,10 +27,9 @@ class attribute.
         upper_limit: Float
```
Jammy2211, line 27: Typing in __init__ not docstring.

Jammy2211 (approved):

PR #602, opened 2022-11-23
timer now handles days clause correctly

Jammy2211:


autofit/non_linear/timer.py
```diff
@@ -46,7 +46,7 @@ def update(self):
         """
 
         try:
-            execution_time = str(dt.timedelta(seconds=time.time() - float(self.start_time)))
+            execution_time = str(dt.timedelta(seconds=time.time() - float(self.start_time)).total_seconds())
```
rhayes777, line 49: Does timedelta need to be used here at all?

rhayes777 (approved):

autofit/non_linear/timer.py
Jammy2211: Apparently not.

PR #604, opened 2022-11-25
Feature/samples

Jammy2211:
Refactoring sampls to remove many redundant methods.


autofit/non_linear/nest/dynesty/abstract.py
```diff
@@ -360,7 +361,7 @@ def sampler_from(
     ):
         raise NotImplementedError()
 
-    def check_pool(self, uses_pool : bool, pool: Pool):
+    def check_pool(self, uses_pool : bool, pool: "Pool"):
```
rhayes777, line 364: Why can't the type be imported here?

autofit/non_linear/samples/samples.py
```diff
@@ -12,7 +14,240 @@
...
+
+    Returns
+    -------
+        A function that returns a 2D image.
+    """
+
+    @wraps(func)
+    def wrapper(
+        obj,
+        sample_index,
```
rhayes777, line 149: sample_index can be implicitly contianed in args or kwargs.

rhayes777: Do we really need to pass as_instance to the decorated function?

rhayes777 (approved):

autofit/non_linear/samples/samples.py
Jammy2211: Lets dicuss next week.

PR #638, opened 2022-12-12
feature/time series

rhayes777:
Implements a generic abstract timeseries and linear concrete implementation.

```python
# Each instance in the time_series has an attribute 't'. Instances here are ModelInstances (for example created using an Aggregator)
time_series = af.LinearTimeSeries([instance_1, instance_2, instance_3])

# We can now create an instance which has a value of t = 3.5 by interpolating
instance = time_series[time_series.t == 3.5)

# We can also interpolate on any arbitrary attribute of the instance
instance = time_series[time_series.some.arbitrary.attribute == -1.0]


Jammy2211 (approved):
Make generic instead of time.

Issue #643, opened 2022-12-18
modify_before_fit not called in summed analysis


Jammy2211:
At the start of a fit, the method `modify_before_fit` is called:

```
    def fit(
        self,
        model,
        analysis: "Analysis",
        info=None,
        pickle_files=None,
        log_likelihood_cap=None,
    ) -> Union["Result", List["Result"]]:

        self.check_model(model=model)

        self.logger.info("Starting search")

        model = analysis.modify_model(model)
        self.paths.model = model
        self.paths.unique_tag = self.unique_tag
        self.paths.restore()

        analysis = analysis.modify_before_fit(paths=self.paths, model=model)
```

This is not happening for a `CombinedAnalysis`, e.g. an example from the autocti_workspace:

https://github.com/Jammy2211/autocti_workspace/blob/main/scripts/imaging_ci/modeling/parallel_x2.py

PR #647, opened 2022-12-19
Combined modify before fit

rhayes777:
Resolvs #643. Also tries to ensure correct behaviour for modifying after fit and modifying models; this may result in bugs in the build


autofit/non_linear/analysis/combined.py
```diff
@@ -42,6 +42,24 @@ def __init__(self, *analyses: Analysis):
...
+    def __getitem__(self, item):
+        return self.analyses[item]
+
+    def modify_before_fit(self, paths: AbstractPaths, model: AbstractPriorModel):
+        """
+        Modify the analysis before fitting.
+
+        Parameters
+        ----------
+        paths
```
Jammy2211, line 54: Good bot

Jammy2211 (approved):

PR #656, opened 2022-12-20
Feature/dynesty plot exceptions

Jammy2211:
Adds try / except clauses for certain ``dynesty`` plots, which may raise an error if the plot occurs too early in a model-fit when the posterior is not yet estimated accurately.

### DIFF
--- a/autofit/non_linear/nest/dynesty/plotter.py
+++ b/autofit/non_linear/nest/dynesty/plotter.py
@@ -10,6 +10,19 @@ class DynestyPlotter(SamplesPlotter):
     
     @staticmethod
     def log_plot_exception(plot_name : str):
+        """
+        Plotting the results of a ``dynesty`` model-fit before they have converged on an
+        accurate estimate of the posterior can lead the ``dynesty`` plotting routines
+        to raise a ``ValueError``.
+
+        This exception is caught in each of the plotting methods below, and this
+        function is used to log the behaviour.
+
+        Parameters
+        ----------
+        plot_name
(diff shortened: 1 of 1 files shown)


autofit/non_linear/nest/dynesty/plotter.py
```diff
@@ -1,13 +1,45 @@
...
+        This exception is caught in each of the plotting methods below, and this
+        function is used to log the behaviour.
+
+        Parameters
+        ----------
+        plot_name
+            The name of the ``dynesty`` plot which raised a ``ValueError``
+        """
+
+        logger.info(
```
rhayes777, line 27: I'd be tempted to make this a warning

autofit/non_linear/nest/dynesty/plotter.py
```diff
@@ -53,32 +106,50 @@ def cornerpoints(self, **kwargs):
...
+
+            self.output.to_figure(structure=None, auto_filename="runplot")
+            self.close()
+
         except ValueError:
-            pass
 
-        self.output.to_figure(structure=None, auto_filename="runplot")
-        self.close()
+            self.log_plot_exception(plot_name="runplot")
```
rhayes777, line 134: You could also do this with a decorator

```python
def log_value_error(func):
    @functools.wraps(func)
    def wrapper(self, *args, **kwargs):
        try:
            return func(self, *args, **kwargs)
        except ValueError:
            self.log_plot_exception(func.__name__)

rhayes777 (approved):

PR #657, opened 2022-12-20
Feature/nuclear mode

Jammy2211:
When multiple model-fits are performed using the same `path_prefix` and `name`,
        the results are populated in the same folder with different unique identifiers.

        By accident, one may perform runs where additional results are placed
        in these folders which are not wanted for the subsequent analysis.

        Removing these results from the directory can be cumbersome, as determining
        the unwanted results based on their unique identifier requires visually inspecting
        them.

        These unwanted results can also make manipulating the results via the database
        problematic, as one may need to again filter based on unique identifier.

        When a run is performed in nuclear mode, all results in every folder are
        deleted except the results corresponding to the unique identifier of that run.

        Therefore, provided the user is 100% certain that the run corresponds to the
        results they want to keep, nuclear mode can be used to remove all unwanted results.

        For example, suppose a folder has 5 results, 4 of which are unwanted and 1 which is
        wanted. If nuclear mode runs, and the model-fit is set up correctly such that the
        identifier created corresponds to the wanted result, all 4 unwanted results
        will be deleted.

        To enable nuclear mode, set the environment variable ``PYAUTOFIT_NUCLEAR_MODE=1``.

        Nuclear model is dangerous, and must be used with CAUTION AND CARE!

### DIFF
--- a/autofit/non_linear/abstract_search.py
+++ b/autofit/non_linear/abstract_search.py
@@ -452,6 +452,7 @@ class NonLinearSearch(AbstractFactorOptimiser, ABC):
         info=None,
         pickle_files=None,
         log_likelihood_cap=None,
+        bypass_nuclear : bool = False
     ) -> Union["Result", List["Result"]]:
         """
         Fit a model, M with some function f that takes instances of the
@@ -570,7 +571,9 @@ class NonLinearSearch(AbstractFactorOptimiser, ABC):
 
         self.logger.info("Removing zip file")
         self.paths.zip_remove()
-        self.paths.zip_remove_nuclear()
+
+        if not bypass_nuclear:
+            self.paths.zip_remove_nuclear()
 
         return result
 
--- a/autofit/non_linear/abstract_search.py
+++ b/autofit/non_linear/abstract_search.py
@@ -85,7 +85,7 @@ class NonLinearSearch(AbstractFactorOptimiser, ABC):
         searches.
 
         Parameters
-        -----------
+    ----------
         name
             The name of the search, controlling the last folder results are output.
         path_prefix
--- a/autofit/non_linear/abstract_search.py
+++ b/autofit/non_linear/abstract_search.py
@@ -571,6 +571,7 @@ class NonLinearSearch(AbstractFactorOptimiser, ABC):
         self.logger.info("Removing zip file")
         self.paths.zip_remove()
         self.paths.zip_remove_nuclear()
+
         return result
 
     @abstractmethod
--- a/autofit/non_linear/abstract_search.py
+++ b/autofit/non_linear/abstract_search.py
@@ -452,7 +452,7 @@ class NonLinearSearch(AbstractFactorOptimiser, ABC):
         info=None,
         pickle_files=None,
         log_likelihood_cap=None,
-        bypass_nuclear : bool = False
+        bypass_nuclear_if_on : bool = False
     ) -> Union["Result", List["Result"]]:
         """
         Fit a model, M with some function f that takes instances of the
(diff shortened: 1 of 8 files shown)


autofit/non_linear/abstract_search.py
```diff
@@ -452,6 +452,7 @@ def fit(
         info=None,
         pickle_files=None,
         log_likelihood_cap=None,
+        bypass_nuclear : bool = False
```
rhayes777, line 455: Should nuclear really be the default?

autofit/non_linear/nest/dynesty/plotter.py
```diff
@@ -1,13 +1,45 @@
...
+        This exception is caught in each of the plotting methods below, and this
+        function is used to log the behaviour.
+
+        Parameters
+        ----------
+        plot_name
+            The name of the ``dynesty`` plot which raised a ``ValueError``
+        """
+
+        logger.info(
```
rhayes777, line 27: Maybe make this a warning or throw an exception?

autofit/non_linear/paths/abstract.py
```diff
@@ -244,6 +244,60 @@ def _zip(self):
...
+        results they want to keep, nuclear mode can be used to remove all unwanted results.
+
+        For example, suppose a folder has 5 results, 4 of which are unwanted and 1 which is
+        wanted. If nuclear mode runs, and the model-fit is set up correctly such that the
+        identifier created corresponds to the wanted result, all 4 unwanted results
+        will be deleted.
+
+        To enable nuclear mode, set the environment variable ``PYAUTOFIT_NUCLEAR_MODE=1``.
+
+        Nuclear model is dangerous, and must be used with CAUTION AND CARE!
```
rhayes777, line 275: typo

autofit/non_linear/paths/abstract.py
```diff
@@ -244,6 +244,60 @@ def _zip(self):
...
+
+        Therefore, provided the user is 100% certain that the run corresponds to the
+        results they want to keep, nuclear mode can be used to remove all unwanted results.
+
+        For example, suppose a folder has 5 results, 4 of which are unwanted and 1 which is
+        wanted. If nuclear mode runs, and the model-fit is set up correctly such that the
+        identifier created corresponds to the wanted result, all 4 unwanted results
+        will be deleted.
+
+        To enable nuclear mode, set the environment variable ``PYAUTOFIT_NUCLEAR_MODE=1``.
```
rhayes777, line 273: If it's off by default then the default value for boolean arguments to functions should reflect this

autofit/non_linear/paths/abstract.py
```diff
@@ -244,6 +244,60 @@ def _zip(self):
...
+                    os.remove(file_to_remove)
+                    logger.info(f"NUCLEAR MODE -- Removed {file_to_remove}")
+                except (IsADirectoryError, FileNotFoundError):
+                    pass
+
+                try:
+                    shutil.rmtree(file_to_remove)
+                    logger.info(f"NUCLEAR MODE -- Removed {file_to_remove}")
+                except (NotADirectoryError, FileNotFoundError):
+                    pass
```
rhayes777, line 299: This could probably be written more elegantly using pathlib

rhayes777 (approved):

rhayes777:
Looks like you need to add a unit test to make the build pass :)

autofit/non_linear/abstract_search.py
Jammy2211: Nuclear mode also requires that the enviroment variable "PYAUTOFIT_NUCLEAR_MODE" = 1.

This is a switch to not use it when its on, which was useful for some reason.

I have added a doc.

autofit/non_linear/nest/dynesty/plotter.py
Jammy2211: Its expected behaviour (Dynesty cant visualize certain things mid-fit) so its better as info.

PR #708, opened 2023-05-10
feature/sub analysis aggregator

### LINKED ISSUE #705: Database support for CombinedAnalysis
The following example script shows that general database functionality works for a normal Analysis object:

https://github.com/Jammy2211/autofit_workspace_test/blob/main/database/directory/general.py

The following script shows that for a combined analysis, it does not work.

https://github.com/Jammy2211/autofit_workspace_test/blob/main/database/directory/general.py

**save_attributes_for_aggregator**

The function `save_attributes_for_aggregator` is not called for a `CombinedAnalysis` and therefore the custom `.pickle` files are not output (e.g. `data.pickle`). 

The main functionality that needs to be implemented is therefore associated with this comment at the end of the script:

```
"""
__Data__

Loading data via the aggregator, to ensure it is output by the model-fit in pickle files and loadable.

In the `general.py` example this provides a list of `data` objects for every model-fit performed (in this example a 
list with a single entry for the single model-fit performed).

For a `CombinedAnalysis` this should provide a list of `data` objects for every model-fit performed, but where
every entry in the list is a list of `data` objects corresponding to each `data` used in each individual `Analysis`.

There is an example in the comment below.
"""
def _data_from(fit: af.Fit):

    data = fit.value(name="data")

    return data

data_gen = agg.map(func=_data_from)

print("Data via Data Gen:")
print([data for data in data_gen])

# If 5 model-fits are performed using CombinedAnalysis, each with 3 datasets, this should return the second data '
# of the 4th model fit.

# print(data_gen[3][1])
```

rhayes777:
Resolves #705.

Values from child analyses can be obtained using the `child_values` method
```python
aggregator.child_values('data')
```

Fits which are not top-level are filtered out during aggregation. These fits can be included by setting a flag.
```python
aggregator = af.Aggregator.from_database('filename', top_level_only=False)

### CODE REVIEW

**Comment by Jammy2211** [comment]
This seems to work:

```
def _data_from(fit: af.Fit):

    data = fit.value(name="data")

    return data

data_gen = agg.child_values(name="data")

data = [data for data in data_gen]
```

Is there any way to do the `child_values` in the generator, so something like:

```
def _data_from(fit: af.Fit):

    data = fit.child_values(name="data")

    return data

data_gen = agg.values(name="data")

data = [data for data in data_gen]
```

I probably doesn't matter but would make it easier to maintain a consistent API in places.

PR #709, opened 2023-05-10
Features/nested

matthewghgriffiths:
Updating `Factor` api to allow nested arguments to factor

See example, 

```
>>> import autofit.graphical as graph 
>>> def func(a, b):
...     a0 = a[0]
...     c = a[1]['c']
...     return a0 * c * b
... 
>>> a, b, c = graph.variables("a, b, c")
>>> func((1, {'c': 2}), 3)
6
>>> factor = graph.Factor(func, [a, {'c': c}], b)
>>> factor({a: 1, b: 3, c:2})
FactorValue(6, {})
>>> factor.numerical_func_jacobian({a: 1, b: 3, c:2})
(FactorValue(6., {}), JacobianVectorProduct((a, c, b) → ∂(FactorValue)ᵀ (a, c, b)))
>>> factor.func_gradient({a: 1, b: 3, c:2})
(FactorValue(6., {}), VariableData({<class 'autofit.mapper.variable.FactorValue'>: 1.0, Variable(a): 5.999999963535174, Variable(c): 2.999999892949745, Variable(b): 1.999999899027216}))
```

If we end up merging functionality with `jax.PyTree` this will allow us to pass objects to functions as well. I think the [jaxtyping](https://github.com/google/jaxtyping) library could also allow closer integration with the declarative PyAutoFit style with jax arrays.


autofit/graphical/factor_graphs/abstract.py
```diff
@@ -293,7 +301,9 @@ def _numerical_factor_jacobian(
                     x_i -= eps
                     indexes = (Ellipsis,) + it.multi_index
                     for v0, jac_v0v_i in jac_v1_i.items():
-                        jac[v0][i][indexes] = jac_v0v_i
+                        key_path = (v0, *ks, indexes)
+                        nested_set(jac, key_path, jac_v0v_i)
+                        # jac[v0][i][indexes] = jac_v0v_i
```
rhayes777, line 306: Commented code

test_autofit/graphical/functionality/test_factor_graph.py
```diff
@@ -77,6 +77,41 @@ def test_factor_jacobian():
...
+
+    assert factor(values) == pytest.approx(f)
+
+    fval, grad = factor.func_gradient(values)
+
+    assert fval == pytest.approx(f)
+    assert grad[a] == pytest.approx(6)
+    assert grad[b] == pytest.approx(2)
+    assert grad[c] == pytest.approx(3)
+
```
rhayes777, line 113: I'd break this down into two tests

rhayes777 (approved):
It's probably worth writing some unit tests for the nested_* functions checking both standard behaviour and some edge cases.

matthewghgriffiths:
If we're merging functionality with jax, we might want to make the `nested_*` functionality align with the functionality in `jax.tree_utils` ?

rhayes777:
I guess at the moment JAX isn't a requirement whereas it would be if we used that functionality

matthewghgriffiths:
Yeah I imagine that we'd want a backstop implementation that doesn't rely on jax but matches it's behaviour (might be a good way of testing the functionality as well)

autofit/graphical/utils.py
```diff
@@ -54,6 +54,98 @@ def is_iterable(arg):
...
+    -------
+    >>> list(nested_items([1, (2, 3), [3, {'a': 1, 'b': 2}]]))
+    [((0,), 1), ((1, 0), 2), ((1, 1), 3), ((2, 0), 3), ((2, 1, 'a'), 1), ((2, 1, 'b'), 2)]
+
+    >>> list(nested_items([1, (2, 3)], [4, (5, 6)]))
+    [((0,), 1, 4), ((1, 0), 2, 5), ((1, 1), 3, 6)]
+    """
+    out, *_ = args
+    if isinstance(out, dict):
+        for k in out:
```
matthewghgriffiths, line 115: to match `jax` functionality (and to provide stable ordering) may want `from k in sorted(out)` @rhayes777

rhayes777: Makes sense

matthewghgriffiths:
Have updated library and added more tests. @rhayes777

PR #711, opened 2023-06-01
Sensitivity Mapping Simulation Visualization and Info

Jammy2211:
The `Simulate` object used in sensitivity mapping now receives a `simulate_path` which is a folder where simulation specific visualization and info can be output.

In the example below, I use this to output the dataset visuals, tracer visuals and a .json containing the tracer used to simulate the data:

class SimulateImaging:
    def __init__(self, mask, psf):

        self.mask = mask
        self.psf = psf

    def __call__(self, instance, simulate_path):
        """
        Set up the `Tracer` which is used to simulate the strong lens imaging, which may include the subhalo in
        addition to the lens and source galaxy.
        """
        tracer = al.Tracer.from_galaxies(
            galaxies=[
                instance.galaxies.lens,
                instance.perturbation,
                instance.galaxies.source,
            ]
        )

        """
        Set up the grid, PSF and simulator settings used to simulate imaging of the strong lens. These should be tuned to
        match the S/N and noise properties of the observed data you are performing sensitivity mapping on.
        """
        grid = al.Grid2DIterate.uniform(
            shape_native=self.mask.shape_native,
            pixel_scales=self.mask.pixel_scales,
            fractional_accuracy=0.9999,
            sub_steps=[2, 4, 8, 16, 24],
        )

        simulator = al.SimulatorImaging(
            exposure_time=300.0,
            psf=self.psf,
            background_sky_level=0.1,
            add_poisson_noise=True,
        )

        dataset = simulator.via_tracer_from(tracer=tracer, grid=grid)
        dataset.apply_mask(mask=self.mask)

        self.output_info(simulate_path=simulate_path, dataset=dataset, tracer=tracer)

        """
        The data generated by the simulate function is that which is fitted, so we should apply the mask for 
        the analysis here before we return the simulated data.
        """
        return dataset

    def output_info(self, simulate_path, dataset, tracer):

        mat_plot = aplt.MatPlot2D(
            output = aplt.Output(path=simulate_path, format="png")
        )

        dataset_plotter = aplt.ImagingPlotter(
            imaging=dataset, mat_plot_2d=mat_plot
        )
        dataset_plotter.subplot_dataset()

        tracer_plotter = aplt.TracerPlotter(
            tracer=tracer, grid=dataset.grid, mat_plot_2d=mat_plot
        )
        tracer_plotter.subplot_lensed_images()

        with open(os.path.join(simulate_path, "tracer.json"), "w+") as f:
            json.dump(tracer.dict(), f, indent=4)

### DIFF
--- a/autofit/non_linear/grid/sensitivity.py
+++ b/autofit/non_linear/grid/sensitivity.py
@@ -459,6 +459,6 @@ class AnalysisFactory:
     def __call__(self):
         dataset = self.simulate_function(
             instance=self.instance,
-            paths=self.paths
+            simulate_path=self.paths.image_path.replace("image", "simulate")
         )
         return self.analysis_class(dataset)
--- a/autofit/non_linear/grid/sensitivity.py
+++ b/autofit/non_linear/grid/sensitivity.py
@@ -432,6 +432,7 @@ class Sensitivity:
                     instance=instance,
                     simulate_function=self.simulate_function,
                     analysis_class=self.analysis_class,
+                    paths=search.paths,
                 ),
                 model=self.model,
                 perturbation_model=perturbation_model,
@@ -444,7 +445,7 @@ class Sensitivity:
 
 class AnalysisFactory:
     def __init__(
-        self, instance, simulate_function, analysis_class,
+        self, instance, simulate_function, analysis_class, paths,
     ):
         """
         Callable to delay simulation such that it is performed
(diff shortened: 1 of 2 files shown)


rhayes777 (approved):
I feel like this breaks separation of concerns

Jammy2211:
> I feel like this breaks separation of concerns

I agree, but the simulation function is user-written and the quantities which are visualized depend on it, so I dont see an obvious way to avoid it.

rhayes777:
> > I feel like this breaks separation of concerns
> 
> I agree, but the simulation function is user-written and the quantities which are visualized depend on it, so I dont see an obvious way to avoid it.

Fair enough

PR #712, opened 2023-06-01
Features/exact

matthewghgriffiths:
Updating graphical api to allow better exact updates. 

Added graphical/regression/test_exact.py to test exact EP updates. 

In that case an exact linear regression and exact probit update is used to build a linear classification model.

### DIFF
--- a/autofit/graphical/expectation_propagation/ep_mean_field.py
+++ b/autofit/graphical/expectation_propagation/ep_mean_field.py
@@ -215,7 +215,6 @@ class EPMeanField(FactorGraph):
             for v in factor_dist.all_variables
         }).prod(
             *factor_mean_field.values(),
-            default=factor_dist.zeros_like(),
         )
 
         model_dist = factor_dist.prod(cavity_dist, default=factor)
--- a/autofit/graphical/expectation_propagation/ep_mean_field.py
+++ b/autofit/graphical/expectation_propagation/ep_mean_field.py
@@ -215,7 +215,7 @@ class EPMeanField(FactorGraph):
             for v in factor_dist.all_variables
         }).prod(
             *factor_mean_field.values(),
-            default=factor_dist,
+            default=factor_dist.zeros_like(),
         )
 
         model_dist = factor_dist.prod(cavity_dist, default=factor)
--- a/autofit/graphical/expectation_propagation/factor_optimiser.py
+++ b/autofit/graphical/expectation_propagation/factor_optimiser.py
@@ -32,6 +32,12 @@ class AbstractFactorOptimiser(ABC):
     ) -> Tuple[MeanField, Status]:
         factor = factor_approx.factor
         cavity_dist = factor_approx.cavity_dist.copy()
+
+        """For exact_fits of FactorApproximations we need the full cavity_dist, 
+        so in this case we need to create zeros_like versions of the missing 
+        cavity_dists - e.g. for the Normal Distribution the zeros_like dist 
+        has infinite variance, note that its logpdf will be - inf 
+        so we don't want to evaluate it in general."""
         for v in factor_approx.mean_field.keys() - cavity_dist:
             cavity_dist[v] = factor_approx.mean_field[v].zeros_like()
 
--- a/autofit/graphical/expectation_propagation/factor_optimiser.py
+++ b/autofit/graphical/expectation_propagation/factor_optimiser.py
@@ -31,7 +31,10 @@ class AbstractFactorOptimiser(ABC):
         self, factor_approx: FactorApproximation, status: Status = Status()
     ) -> Tuple[MeanField, Status]:
         factor = factor_approx.factor
-        cavity_dist = factor_approx.cavity_dist
+        cavity_dist = factor_approx.cavity_dist.copy()
+        for v in factor_approx.mean_field.keys() - cavity_dist:
+            cavity_dist[v] = factor_approx.mean_field[v].zeros_like()
+
         with LogWarnings(logger=self.logger, action="always") as caught_warnings:
             if factor._calc_exact_update:
                 factor_mean_field = factor.calc_exact_update(cavity_dist)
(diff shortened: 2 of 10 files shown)


autofit/graphical/expectation_propagation/ep_mean_field.py
```diff
@@ -215,7 +215,6 @@ def factor_approximation(self, factor: Factor) -> FactorApproximation:
             for v in factor_dist.all_variables
         }).prod(
             *factor_mean_field.values(),
-            default=factor_dist,
```
matthewghgriffiths, line 218: This causes incorrect initialisation of the cavity dist if a variable is used by only one factor.

rhayes777: It's quite common in our usage for variables to only be used by a single factor. Many factors have 'dangling' variables which are not shared

rhayes777: Do you mean you've fixed that issue or newly introduced it?

matthewghgriffiths: I think I fixed it. 

The way that you solved the dangling variable problem was causing variance shrinkage in the case of dangling variables.

autofit/graphical/expectation_propagation/optimiser.py
```diff
@@ -188,15 +186,25 @@ def __init__(
...
+        if paths is None:
+            try:
+                paths = DirectoryPaths(identifier=str(Identifier(factor_graph)))
+            except KeyError: 
+                pass 
+
+        self.paths = paths
+        if self.paths:
+            for optimiser in self.factor_optimisers.values():
+                optimiser.paths = self.paths
```
rhayes777, line 202: We probably want each factor_optimiser to have a different paths to ensure that they do not clash.

If multiple DynestyStatic had the same paths, for example, we might see them loading each other's optimisation state.

You could use the create_child method and pass the name of each factor.

autofit/graphical/mean_field.py
```diff
@@ -89,6 +90,9 @@ def __init__(
             self.log_norm = log_norm
             self._plates = self.sorted_plates if plates is None else plates
 
+    def copy(self) -> "MeanField":
+        return type(self)(self)
```
rhayes777, line 94: A type of self of self!

autofit/graphical/expectation_propagation/ep_mean_field.py
matthewghgriffiths: This way the dangling variable has no cavity distribution associated with it in the `FactorApproximation` which is fine for a lot of factor optimisations.

If you require access to the cavity distribution as a prior then you shouldn't have a dangling variable. (The prior should be present as a second factor connected to the variable).

autofit/graphical/mean_field.py
```diff
@@ -212,44 +220,47 @@ def rescale(self, rescale: Dict[Variable, float]) -> "MeanField":
...
 
     @property
     def std(self):
-        return VariableData({v: dist.std for v, dist in self.items()})
+        return self.attr("std")
 
     @property
     def scale(self):
-        return VariableData({v: dist.scale for v, dist in self.items()})
+        return self.attr("scale")
```
rhayes777, line 244: Nice

rhayes777 (approved):
Looks like it only failed because of a drop in code coverage. Is there a test or two you could write? I think the paths stuff is probably not working properly but it would be a bit trickier to write a decent test for that as you would have to mock up the optimisers.

Also some docs would be nice

Otherwise looks good

autofit/graphical/expectation_propagation/factor_optimiser.py
```diff
@@ -31,7 +31,10 @@ def exact_fit(
         self, factor_approx: FactorApproximation, status: Status = Status()
     ) -> Tuple[MeanField, Status]:
         factor = factor_approx.factor
-        cavity_dist = factor_approx.cavity_dist
+        cavity_dist = factor_approx.cavity_dist.copy()
+        for v in factor_approx.mean_field.keys() - cavity_dist:
+            cavity_dist[v] = factor_approx.mean_field[v].zeros_like()
```
matthewghgriffiths, line 36: @rhayes777 

For `exact_fit`s of `FactorApproximation`s we need the full cavity_dist, so in this case we need to create `zeros_like` versions of the missing `cavity_dist`s - e.g. for the Normal Distribution the zeros_like dist has infinite variance, note that its logpdf will be `- inf` so we don't want to evaluate it in general.

rhayes777: Interesting. Could you write something along those lines in the code docs?

autofit/graphical/expectation_propagation/ep_mean_field.py
rhayes777: Actually yeah I guess every variable at least has a PriorFactor attached and so would be shared.

Jammy2211 (approved):

autofit/graphical/expectation_propagation/ep_mean_field.py
matthewghgriffiths: A test for this happening would be to run a factor optimisation multiple times in a row for the same factor, if the variance shrinks each step, then you're doing something like this.

autofit/graphical/expectation_propagation/optimiser.py
matthewghgriffiths: I have no idea what `paths` are supposed to do but it was breaking some of my code, so I put in code paths so that it became optional. In a later commit I've set the paths attribute to be optionally set if it doesn't exist.

matthewghgriffiths:
I don't know why it failed. It works on my machine...

autofit/graphical/expectation_propagation/optimiser.py
matthewghgriffiths: Actually had to revert the change, if you don't have this line `test_autofit/graphical/info/test_output.py` fails.

rhayes777: It's basically an object that handles saving and loading data for optimisations. We used to have path manipulation stuff litered all over the optimiser classes so we put it all in one place. It can also conveniently be swapped out for a database client so data is persisted directly into a SQL database rather than being saved in a directory structure.

PR #721, opened 2023-06-09
feature/covariance interpolator

rhayes777:
First cut of an interpolator that respects covariance. The interpolator must be made using a list of PDFSamples such that covariance information is available.

```python
interpolator = CovarianceInterpolator(pdf_samples)
instance = interpolator(interpolator.t == 1)
print(instance.gaussian.centre)
```

DynestyStatic is used and the relationship between each variable and time is assumed to be linear.

Seems to give correct results but more testing is required.


Jammy2211:
> The interpolator must be made using a list of PDFSamples such that covariance information is available.

Does this use more information than the covariance matrix?

rhayes777:
No just the covariance matrix

Jammy2211:
Ok we'll be putting that out as a separate .csv file soon enough!

autofit/interpolator/covariance.py
```diff
@@ -40,7 +40,7 @@ def __repr__(self):
         return str(self)
 
 
-class LinearAnalysis(Analysis):
+class CovarianceAnalysis(Analysis):
```
Jammy2211, line 43: We're using autofit in autofit!

Jammy2211 (approved):

Jammy2211:
I will set up a CTI example when I have time and add visualization.

autofit/interpolator/covariance.py
rhayes777: Autoception

PR #727, opened 2023-06-12
feature/model json

### LINKED ISSUE #716: Use model.json instead of model.pickle
Currently in PyAutoFit, outputs are stored as .pickle files, which are not ideal for the following reasons:

- They are specific to PyAutoFit and the version of software used for fitting, breaking when imports change.
- They are not human readable and thus difficult to check.

For Euclid CTI calibration, I want as much of our outputs as possible to be .json files, to mitigate these issues.

This issue focuses on turning the `model.pickle` object into a `model.json` file.

The `model.pickle` is created by the following method:

```
    def save_all(
            self,
            search_config_dict=None,
            info=None,
            pickle_files=None
    ):
        search_config_dict = search_config_dict or {}
        info = info or {}
        pickle_files = pickle_files or []

        self.save_identifier()
        self.save_parent_identifier()
        self._save_search(config_dict=search_config_dict)
        self._save_model_info(model=self.model)
        self._save_parameter_names_file(model=self.model)
        self.save_object("info", info)
        self.save_object("search", self.search)
        self.save_object("model", self.model)
        self._save_metadata(
            search_name=type(self.search).__name__.lower()
        )
        self._move_pickle_files(pickle_files=pickle_files)
 ```

Other than loading via the `Aggregator`, it does not have a specific use case (e.g. it is not loaded to resume
results and does not contain samples or other quantities that are important).

I therefore propose that we remove pickling of model objects and have dedicated functionality for each which
can output them to a .json file, and load them again via a .json file. From what I can tell, this should not be
difficult.

This should be saved in a new folder for a fit called `jsons` which will contain all objects of this nature.

Obviously, PyAutoFit may be updated and the .json files may become out of date. Alternatively, the Python classes
used to compose the model may be updated.

Would it be possible to include functionality that when the model is loaded, it checks the .json entries compared to 
the current version of the PyAutoFit search, raises an error explaining why they are different, and shows the user 
how to input the missing entries via a kwarg?

### PR DESCRIPTION
**Pull request by rhayes777**
Resolves #716. The model is now converted into a dictionary which is written to a jsons directory as a JSON or stored as a JSON object in the database.

### DIFF
--- a/autofit/aggregator/search_output.py
+++ b/autofit/aggregator/search_output.py
@@ -1,3 +1,4 @@
+import json
 import logging
 import os
 import pickle
@@ -7,6 +8,7 @@ from pathlib import Path
 import dill
 
 from autofit.non_linear import abstract_search
+from autofit.mapper.prior_model.abstract import AbstractPriorModel
 
 original_create_file_handle = dill._dill._create_filehandle
 
@@ -37,6 +39,10 @@ class Output:
     @property
     def pickle_path(self):
         return self.directory / "pickles"
+
+    @property
+    def json_path(self):
+        return self.directory / "jsons"
 
     def __getattr__(self, item):
         """
@@ -133,10 +139,9 @@ class SearchOutput(Output):
         """
         if self.__model is None:
             try:
-                with open(self.pickle_path / "model.pickle", "r+b") as f:
-                    self.__model = pickle.loads(f.read())
+                with open(self.json_path / "model.json") as f:
+                    self.__model = AbstractPriorModel.from_dict(json.load(f))
             except (FileNotFoundError, ModuleNotFoundError) as e:
-                print(self.pickle_path)
                 logging.exception(e)
         return self.__model
 
--- a/autofit/database/aggregator/scrape.py
+++ b/autofit/database/aggregator/scrape.py
@@ -1,3 +1,4 @@
+import json
 import logging
 import os
 import pickle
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

**Comment by Jammy2211** [comment]
A desired goal is to ensure that as much of the output as possible is backwards compatible.

As described here (https://github.com/rhayes777/PyAutoFit/issues/720), the `result.json` file will look as follows, and thus not be dependent on the model itself (retaining BC):

```
{
    "max_log_likelihood" : { 
        "model.gaussian.centre" : 49.994 (do not round this value like is done for `model.result`),
        "model.gaussian.normalization" : 25.258,
        "model.gaussian.sigma" : 10.115
        }
}
```

The `model.json` file produced by this PR is:

```
{
    "type": "collection",
    "gaussian": {
        "class_path": "autofit.example.model.Gaussian",
        "type": "model",
        "centre": {
            "lower_limit": 0.0,
            "upper_limit": 100.0,
            "type": "Uniform"
        },
        "normalization": {
            "lower_limit": 0.01,
            "upper_limit": 100.0,
            "type": "LogUniform"
        },
        "sigma": {
            "lower_limit": 0.0,
            "upper_limit": Infinity,
            "type": "Gaussian",
            "mean": 10.0,
            "sigma": 5.0
        }
    }
}
```

This is not backwards compatible to changes to the model or autofit.

I think we will ultimately require two `model.json` files:

1) One which links priors in the same model indpendent way as `result.json`:

```
{
    "model.gaussian.centre" {
            "lower_limit": 0.0,
            "upper_limit": 100.0,
            "type": "Uniform"
        },
        "model.gaussian.normalization": {
            "lower_limit": 0.01,
            "upper_limit": 100.0,
            "type": "LogUniform"
        },
        "model.gaussian.sigma": {
            "lower_limit": 0.0,
            "upper_limit": Infinity,
            "type": "Gaussian",
            "mean": 10.0,
            "sigma": 5.0
        }
    }
}
```

2) A `model_mapper.json` which retains all of the other information lost above, for example this stuff:

```
    "type": "collection",
    "gaussian": {
        "class_path": "autofit.example.model.Gaussian",
        "type": "model",
```

This .json will serve the role of doing things like making instances.

If we BC is lost, we can raise an exception when a user uses functionality depedent on `model_mapper.json`, but they can probably still do what they need in the end.

I am not sure if now is the right time to do this split, but I put this here so we know where we are going.

Jammy2211 (approved):

PR #731, opened 2023-06-12
feature/search json

rhayes777:
Adds methods to convert generic objects to and from dictionary representation


rhayes777:
This will facilitate search.json

PR #760, opened 2023-07-10
Feature/samples no pickle

Jammy2211:
This changes the output paths of a model-fit to be more intuitive.

It also removes the output of `samples.pickle`, as we are removing all pickles, and relies on `samples.csv` or internal search results to load results.

There are 3 unit tests broken and the source code interface of searches and samples is a bit of a mess, but I wont have time to clean this up in the near future.

Would be good to get everything merged so we can begin testing the new outputs.


Jammy2211:
Also can you rename the `jsons` folder to `files`.

PR #764, opened 2023-07-16
Feature/samples no pickle fixed

Jammy2211:
This changes the outputs of a search to a more logical and structured set of paths.

It also removes `samples.pickle` and relies on either internal search results (in a .dill file) or the samples .csv file to be used.

This ended up a large refactor of searches and samples, which should make the code cleaner and more maintainable.


autofit/non_linear/search/abstract_search.py
```diff
@@ -884,7 +874,16 @@ def remove_state_files(self):
         pass
 
     def samples_from(self, model):
```
rhayes777, line 876: Might be worth adding docs explaining the different ways samples get loaded

autofit/non_linear/search/mcmc/abstract_mcmc.py
```diff
@@ -43,3 +42,9 @@ def __init__(
     @property
     def config_type(self):
         return conf.instance["non_linear"]["mcmc"]
+
+    def samples_via_csv_from(self, model):
```
rhayes777, line 46: Docs would be good

rhayes777 (approved):

PR #771, opened 2023-07-21
feature/composition

rhayes777:
Integrates the new reference dictionary. 

This can be used when creating an aggregtor:

```python
from autofit.aggregator import Aggregator
from autoconf.class_path import get_class_path
import autofit as af


aggregator = Aggregator(
    "directory",
    reference={"path.to.model": get_class_path(af.Exponential)},
)
```

Or when scraping for a database aggregator:

```python

aggregator = af.Aggregator.from_database("database.sqlite")
aggregator.add_directory(
    "directory",
    reference={"": get_class_path(af.Exponential)},
)
```

The key is a . separated path through the model. An empty string means the top level object.


rhayes777:
This also fixed two other issues:

- Priors were not properly parsed from JSON since implementing linked prior recovery
- Models with GaussianPriors were not correctly persisted in the database

Jammy2211 (approved):

PR #775, opened 2023-07-24
feature/attribute types

rhayes777:
Implements persisting of additional types in the files directory. Ensures types can be persisted as filees and scraped into the database.


Jammy2211 (approved):
process_calibrate_cti.iterati

rhayes777:
> Do we need to Astropy as an optional requirement to autofit?

Ah no we haven't I'll make those imports local only

Jammy2211:
Add it to `optional_requirements.txt` as well:

Same as autoarray, so: `astropy>=5.0`

PR #795, opened 2023-08-14
feature/scrape samples

rhayes777:
Ensure that scraping creates a Samples object when CSV and JSON samples data available in files

### DIFF
--- a/autofit/database/aggregator/scrape.py
+++ b/autofit/database/aggregator/scrape.py
@@ -9,6 +9,7 @@ import numpy as np
 
 from .. import model as m
 from ..sqlalchemy_ import sa
+from ... import Samples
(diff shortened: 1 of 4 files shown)


autofit/database/aggregator/scrape.py
```diff
@@ -11,6 +12,7 @@
 from ..sqlalchemy_ import sa
```
Jammy2211, line 12: Is there a reason to use relative imports here?

Jammy2211 (approved):

autofit/database/aggregator/scrape.py
rhayes777: No mostly likely pycharm put them there automatically

PR #804, opened 2023-09-05
Feature/lh timeout

Jammy2211:
Allows one to time-out a log likelihood function call in order to receive an error message enabling the diagnosis of infinite loops.

By default this is off, but it can be switched on by updating the following entry of the `general.yaml` config file:

```
test:
  lh_timeout_seconds:               # If a float is input, the log_likelihood_function call is timed out after this many seconds, to diagnose infinite loops. Default is None, meaning no timeout.
```

### DIFF
--- a/autofit/non_linear/fitness.py
+++ b/autofit/non_linear/fitness.py
@@ -1,3 +1,4 @@
+import logging
 import numpy as np
 
 from autoconf import conf
@@ -6,10 +7,16 @@ from autofit import exc
 
 from timeout_decorator import timeout
 
-try:
-    timeout_seconds = conf.instance["general"]["test"]["lh_timeout_seconds"]
-except KeyError:
-    timeout_seconds = None
+def get_timeout_seconds():
+
+    try:
+        return conf.instance["general"]["test"]["lh_timeout_seconds"]
+    except KeyError:
+        pass
+
+logger = logging.getLogger(__name__)
+
+timeout_seconds = get_timeout_seconds()
 
 class Fitness:
     def __init__(
--- a/autofit/non_linear/fitness.py
+++ b/autofit/non_linear/fitness.py
@@ -1,6 +1,10 @@
 import numpy as np
 
+from autoconf import conf
+
 from autofit import exc
+
+from timeout_decorator import timeout
 
 class Fitness:
     def __init__(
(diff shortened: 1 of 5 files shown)


autofit/non_linear/search/abstract_search.py
```diff
@@ -597,6 +599,14 @@ def pre_fit_output(
                 model=model,
             )
 
+            timeout_seconds = get_timeout_seconds()
+
+            if timeout_seconds is not None:
+
+                logger.info(f"\n\n ***Log Likelihood Function timeout is "
+                            f"turned on and set to {timeout_seconds} seconds.***\n")
+
```
rhayes777, line 608: Doesn't do anything yet?

autofit/non_linear/fitness.py
```diff
@@ -1,7 +1,23 @@
...
+def get_timeout_seconds():
+
+    try:
+        return conf.instance["general"]["test"]["lh_timeout_seconds"]
+    except KeyError:
+        pass
+
+logger = logging.getLogger(__name__)
+
+timeout_seconds = get_timeout_seconds()
```
rhayes777, line 19: redundant?

autofit/non_linear/fitness.py
```diff
@@ -1,7 +1,23 @@
...
 from autofit import exc
 
+from timeout_decorator import timeout
+
+def get_timeout_seconds():
+
+    try:
+        return conf.instance["general"]["test"]["lh_timeout_seconds"]
+    except KeyError:
+        pass
```
rhayes777, line 15: Catch the key error at the point of use rather than having to add a None-check

rhayes777 (approved):

autofit/non_linear/search/abstract_search.py
Jammy2211: It prints that the time out feature is turned on.

I couldnt do this in `fitness.py` cause it needs to be wtihin the `is_master` parallel calls.

autofit/non_linear/fitness.py
Jammy2211: `get_timeout_seconds()` is used twice in the code, so catching the error at the point of use would lead to duplicaiton.

autofit/non_linear/search/abstract_search.py
rhayes777: Ah makes sense

PR #817, opened 2023-09-19
Feature/test mode dynesty

Jammy2211:
`Nautilus` does not have functionality to input the maximum number of iterations. This means it cannot be used in **PyAutoFit** test mode, where searches are skipped instantly.

Now that all workspaces use `Nautilus` by default, this poses a problem where they cannot be integration tested quickly.

This PR puts an if statement in such that if `Nautilus` is running in test mode, it is swapped out for `Dynesty`.

### DIFF
--- a/autofit/__init__.py
+++ b/autofit/__init__.py
@@ -130,4 +130,4 @@ def save_abc(pickler, obj):
 
 conf.instance.register(__file__)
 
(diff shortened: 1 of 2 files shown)


rhayes777 (approved):
It might be more elegant to do this using the factory pattern in `__new__` method of the Nautilus optimiser

PR #864, opened 2023-10-30
feature/grid search attribute

### LINKED ISSUE #861: Grid Search Result LimitList of parameter
A `GridSearchResult` has a `LimitList` type which provides list of lists of quantities that are mapped over, for example the upper limit of the prior of every grid search parameter:

```
    @property
    def upper_limits_lists(self) -> LimitLists:
        """
        The upper values for each grid square
        """
        return [
            [limit + self.step_size for limit in limits]
            for limits in self.lower_limits_lists
        ]
```

This `LimitList` structure is a key property of the `GridSearchResult`, it is represents the structure of all results performed for the `GridSearch`.

In `PyAutoLens`, I manually map over inferred results to plot the results of a grid search:

```
        instance_list = [
            None if samples.median_pdf() is None else samples.median_pdf()
            for samples in self.grid_search_result_with_subhalo.samples
        ]

        return [
                None if instance is None else instance.galaxies.subhalo.mass.mass_at_200
                for instance in instance_list
            ]
```

A much more concise API should be available, where I simply query the name of a parameter and quantity I want, and the `LimitList` corresponding to it is returned.

```
mass_at_200_list = self.grid_search_result.values(parameter="mass_at_200", quantity=samples.median_pdf())
```

I am not 100% sure on the exact API, but you should get the idea that for a `GridSearchResult` I should be able to get the `LimitList` for any parameter in `GridSearchResult.samples` and should be able to choose what attribute of the `Samples` I want (e.g. `median_pdf`, `max_log_likelihood`).

rhayes777:
Resolves #861

GridSearchResult now has a method called 'attribute_grid' which can be used to obtain a numpy array in the same dimensions as the grid search
```python
result.attribute_grid("centre")
```

This supports querying attributes deeper in the model using . delimited strings (or iterables)

```python
result.attribute_grid("gaussian.centre")
```

Values are taken from the highest log-likelihood instance. Currently SamplesSummary does not retain the median instance.

### DIFF
--- a/autofit/non_linear/grid/grid_search/result.py
+++ b/autofit/non_linear/grid/grid_search/result.py
@@ -184,12 +184,26 @@ class GridSearchResult:
         """
         return self._list_to_native(lst=[sample for sample in self.samples])
 
-    def attribute_grid(self, attribute_name: str):
+    def attribute_grid(self, attribute_name: str) -> np.ndarray:
+        """
+        Get a list of the attribute of the best instance from every search in a numpy array with the native dimensions
+        of the grid search.
+
+        Parameters
+        ----------
+        attribute_name
+            The name of the attribute to get from the instance
+
+        Returns
+        -------
+        A numpy array of the attribute of the best instance from every search in the grid search.
+        """
+
         def _get_attribute(
             obj: Union[np.ndarray, SamplesInterface],
         ):
             if isinstance(obj, np.ndarray):
-                return [_get_attribute(item) for item in obj]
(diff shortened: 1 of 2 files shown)


rhayes777:
It's also worth noting you can get a grid of any arbitrarily part of the model, for example:
```python
result.attribute_grid("gaussian")
```
Could be used to obtain a grid of instances of the Gaussian class

Jammy2211 (approved):

PR #868, opened 2023-11-09
Sensitivity Refactor

Jammy2211:
Previously, sensitivity mapping required input of a non linear search which was used for both model-fits.

However, this had a number of limitations:

1) The way the search was performed could not be very customized.
2) Sensitivity mapping may not even use non-linear searches -- you could imagine simply mapping over a goodness-of-fit metric which is computed in a much simpler way.

I have therefore extracted out the search functionality such that it is now defined by the use. Here is an example of how we would set up the base fit for the previous functionality:

```
class BaseFit:

    def __init__(self, analysis_cls):
        """
        Class used to fit every dataset used for sensitivity mapping with the base model (the model without the 
        perturbed feature sensitivity mapping maps out).

        In this example, the base model therefore does not include the extra Gaussian feature, but the simulated 
        dataset includes one.
        
        The base fit is repeated for every parameter on the sensitivity grid and compared to the perturbed fit. This 
        maps out the sensitivity of every parameter is (e.g. the sensitivity of the normalization of the Gaussian 
        feature).

        The `__init__` constructor can be extended with new inputs which can be used to control how the dataset is
        fitted, below we include an input `analysis_cls` which is the `Analysis` class used to fit the model to the
        dataset.
        
        Parameters
        ----------
        analysis_cls
            The `Analysis` class used to fit the model to the dataset.
        """
        self.analysis_cls = analysis_cls

    def __call__(self, dataset, model, paths):
        """
        The base fitting function which fits every dataset used for sensitivity mapping with the base model.

        This function receives as input each simulated dataset of the sensitivity map and fits it, in order to
        quantify how sensitive the model is to the perturbed feature.

        In this example, a full non-linear search is performed to determine how well the model fits the dataset.
        The `log_evidence` of the fit is returned which acts as the sensitivity map figure of merit.

        Parameters
        ----------
        dataset
            The dataset which is simulated with the perturbed model and which is fitted.
        model
            The model instance which is fitted to the dataset, which does not include the perturbed feature.
        paths
            The `Paths` instance which contains the path to the folder where the results of the fit are written to.
        """

        search = af.DynestyStatic(
            paths= paths.for_sub_analysis(analysis_name="[base]"),
            nlive=25,
            unique_tag=dataset_name,
            session=session,
            iterations_per_update=50000,
        )

        analysis = self.analysis_cls(dataset=dataset)

        return search.fit(
            model=model, analysis=analysis
        )
```

There is one issue, that the following code:

```
            paths = self.paths.for_sub_analysis(
                label,
            )
```

Is not extending the paths with the `label` correctly. Previously, this worked on `search.paths` so I guess its to do with that swap. I think this is leading to a broken unit test.

The issue is see clearly because when we run the following integration test:

https://github.com/Jammy2211/autofit_workspace_test/blob/main/scripts/database/scrape/sensitivity.py

The output paths do not include the parameter name `e.g. normalization_75.000` or the  `[base]` and `[perturbed]` tags.

image


autofit/non_linear/grid/sensitivity.py
```diff
@@ -4,100 +4,88 @@
...
-        self.search = search.copy_with_paths(
-            search.paths.for_sub_analysis(
-                "[base]",
-            )
-        )
-        self.perturbed_search = search.copy_with_paths(
-            search.paths.for_sub_analysis(
-                "[perturbed]",
-            )
-        )
```
rhayes777, line 96: This is one part of your problem. I guess you need to manually update this in your 'Fit' classes

autofit/non_linear/grid/sensitivity.py
```diff
@@ -386,42 +374,9 @@ def _perturbation_models(self) -> Generator[AbstractPriorModel, None, None]:
...
-        Returns
-        -------
-        A non linear search, copied from the instance search
-        """
-        paths = self.search.paths
-        search_instance = self.search.copy_with_paths(
-            paths.for_sub_analysis(
-                name_path,
-            )
-        )
```
rhayes777, line 422: And here is the other part

rhayes777 (approved):
I'm not really sure what I'm supposed to do here. The change removes the functionality responsible for modifying search paths and it's not really obvious how it would be put back in. 

Much of what the current sensitivity code does it managing those optimisations. You probably want to extract a common Sensitivity interface which the existing Sensitivity imlpements and that you can then also implement to create a different but similar system that is not dependent on optimising each grid square.

rhayes777:
OK nvm I'm just sleep deprived. I think the issue is resolved now

PR #875, opened 2023-11-17
feature/database info

### LINKED ISSUE #770: Add database.info file
Once we have a database.sqlite file, we have no understanding of its contents. Investigating its contents requires 
us to probe it via .sqlite.

A `database.sqlite` file should have an associated `database.info` file, which lists the contents of the
dataabse in an concise and informative way.

I would suggest a .csv with the following format:

```
entry number                         unique_id                           name                    unique_tag                     total_free_parameters                       finished
           0  a14161122b532586a68b69acdf10895c                 search_name[1]         dataset_tag_example_0                                         3                           True
           1  erfopiu7n9837nr9e9fnef0980e8fn0e                 search_name[1]         dataset_tag_example_1                                         3                           True
```

If we build a database from an output folder via `scrape` this info file should be created.

If the database is created via a session this info file should constantly be being updated.

rhayes777:
Resolves #770

Whenever database session is committed info is also written describing each fit in the database

### DIFF
--- a/test_autofit/aggregator/test_reference.py
+++ b/test_autofit/aggregator/test_reference.py
@@ -52,9 +52,19 @@ def test_database(database_aggregator):
     assert model.cls is af.Exponential
 
 
-def test_info(database_aggregator):
-    info = Info(database_aggregator.session)
+@pytest.fixture(name="info")
+def make_info(database_aggregator):
+    return Info(database_aggregator.session)
+
+
+def test_query_fits(info):
+    fits = info.fits
     assert len(info.fits) == 3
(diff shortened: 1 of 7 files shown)


test_autofit/aggregator/test_reference.py
```diff
@@ -71,4 +71,12 @@ def test_database_info(
     database_aggregator,
     output_directory,
 ):
-    assert (output_directory / "database.info").exists()
+    print((output_directory / "database.info").read_text())
```
Jammy2211: ?

Jammy2211 (approved):

PR #877, opened 2023-11-17
print message when lots of searches are aggregated form the directory

rhayes777:


autofit/aggregator/aggregator.py
```diff
@@ -115,6 +115,10 @@ def __init__(
         search_outputs
             A list of search_outputs
         """
+        if len(search_outputs) > 20:
+            print(
+                "When aggregating many searches it can be more efficient to use the database."
```
Jammy2211, line 120: Add following text to message:

"Checkout the database cookbook at this URL: https://pyautofit.readthedocs.io/en/latest/cookbooks/database.html"

Jammy2211 (approved):

PR #878, opened 2023-11-17
fix how nautilus test is setup to use dynesty paths

Jammy2211:
Simple fix to how paths are set up when switching from `Nautilus` to `Dynesty` in test mode.

Hopefully `Nautilus` will support a fixed number of iteraitons soon, meaning this code can be deleted.

### DIFF
--- a/autofit/non_linear/search/abstract_search.py
+++ b/autofit/non_linear/search/abstract_search.py
@@ -530,9 +530,7 @@ class NonLinearSearch(AbstractFactorOptimiser, ABC):
                 from pathlib import Path
 
                 search = DynestyStatic(
-                    name=self.name,
-                    unique_tag=self.unique_tag,
-                    path_prefix=Path(self.paths.path_prefix),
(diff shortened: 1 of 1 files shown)


autofit/non_linear/search/abstract_search.py
```diff
@@ -530,9 +530,7 @@ class represented by model M and gives a score for their fitness.
                 from pathlib import Path
 
                 search = DynestyStatic(
-                    name=self.name,
-                    unique_tag=self.unique_tag,
-                    path_prefix=Path(self.paths.path_prefix),
```
rhayes777, line 535: ?

rhayes777 (approved):

autofit/non_linear/search/abstract_search.py
Jammy2211: I'm having to do weird things to get test mode to work with `Nautilus`, but this basically ensures the paths are used correclty.

PR #879, opened 2023-11-17
feature/configure output

rhayes777:
- fix test
- remove unused method
- lint fixes
- integrate conditional output

### DIFF
--- a/autofit/config/output.yaml
+++ b/autofit/config/output.yaml
@@ -1,0 +1,1 @@
+default: true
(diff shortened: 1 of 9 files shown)


autofit/config/output.yaml
```diff
@@ -0,0 +1 @@
+default: true
```
Jammy2211, line 1: Add description to config file, like here:

https://github.com/Jammy2211/autofit_workspace/blob/release/config/notation.yaml

Add comments to each entry like here:

https://github.com/Jammy2211/autofit_workspace/blob/release/config/general.yaml

Jammy2211 (approved):

PR #880, opened 2023-11-19
Feature/model bounded

Jammy2211:
A new way to pass models and priors for non-linear search chaining, to complement `model_absolute` and `model_relative`.

The new method, `model_bounded`, passes the results as `UniformPriors` centred on a result's maximum likelihood values and with a `lower_limit` and `upper_limit` which have had the input bound value `b` subtracted and added to them.

### DIFF
--- a/autofit/non_linear/grid/sensitivity.py
+++ b/autofit/non_linear/grid/sensitivity.py
@@ -109,7 +109,7 @@ class Job(AbstractJob):
         )
 
         perturb_model = copy(self.model)
-        perturb_model.perturbation = self.perturb_model
+        perturb_model.perturb = self.perturb_model
 
         perturb_result = self.perturb_fit_cls(
             model=perturb_model,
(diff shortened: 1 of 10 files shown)


autofit/non_linear/grid/sensitivity.py
```diff
@@ -396,6 +399,13 @@ def _make_jobs(self) -> Generator[Job, None, None]:
         for number, (perturb_instance, perturb_model, label) in enumerate(
             zip(self._perturb_instances, self._perturb_models, self._labels)
         ):
+
+            if self.perturb_model_prior_func is not None:
```
rhayes777, line 403: You could also make this a lambda function that does nothing by default.

```python
self. perturb_model_prior_func = lambda perturb_instance, perturb_model: perturb_model

rhayes777 (approved):

PR #903, opened 2024-01-15
update resample criteria

Jammy2211:
Add second critieria to initializer resample crtieria, rejecting models where the likelihood is below -1e98 (e.g. clearly a wrong sample).


rhayes777 (approved):
lol

PR #923, opened 2024-02-05
feature/visualise graph

rhayes777:
Visualise models as graphs

- Graphs are output to model_graph.html
- Collections and Models are square nodes
- Priors are circular nodes
- Attributes are represented by vertices between nodes
- Nodes are given unique colours per type


rhayes777:
Screenshot 2024-02-05 at 16 08 19

Jammy2211 (approved):

PR #940, opened 2024-02-17
Feature/prior passer simplify

Jammy2211:
Prior passing used for non-linear search chaining decided the size of the `GaussianPrior` sigma values from two quantities:

- The size of the sigma specified in config files for each parameter.
- The error inferred from the previous model-fit.

I have decided that we do not need to use the errors inferred from a previous model-fit to do prior linking, with the main motivations being:

- It is nasty using results of a previous fit to set the priors of a subsequent fit.
- It adds code complexity, as we have to have certain results in a `Samples` object available to do this linking.

This PR removes using errors to do prior passing, simplifying the source code.

To enable prior passing, the code used to use a `gaussians_tuples` variable, which has now been simplified to a `prior_means` variable.

### DIFF
--- a/autofit/mapper/prior_model/abstract.py
+++ b/autofit/mapper/prior_model/abstract.py
@@ -906,7 +906,7 @@ class AbstractPriorModel(AbstractModel):
         raise NotImplementedError()
 
     def mapper_from_gaussian_tuples(
-        self, tuples, a=None, r=None, use_errors=True, use_widths=True, no_limits=False
+        self, tuples, a=None, r=None, use_widths=True, no_limits=False
     ):
         """
         The widths of the new priors are taken from the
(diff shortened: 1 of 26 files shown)


autofit/mapper/prior_model/abstract.py
```diff
@@ -966,9 +959,9 @@ def mapper_from_gaussian_tuples(
             if a is not None:
                 width = a
             elif r is not None:
-                width = r * mean
+                width = r * means[i]
```
rhayes777, line 962: Use zip instead of enumerate to get the mean without indexing

autofit/non_linear/paths/directory.py
```diff
@@ -206,6 +208,18 @@ def load_search_internal(self):
...
+        """
+        Remove the internal representation of a non-linear search.
+
+        This deletes the entire `search_internal` folder, including a .pickle / .dill file containing the interal
+        results and files with the timer values.
+
+        This folder can often have a large filesize, thus deleting it can reduce hard-disk use of the model-fit.
+        """
+        shutil.rmtree(self.search_internal_path)
+        os.rmdir(self.search_internal_path)
```
rhayes777, line 221: rmtree removes the top level directory so os.rmdir will throw a FileNotFoundError

rhayes777 (approved):

PR #967, opened 2024-03-14
Feature/samples filter

Jammy2211:
For large model fitting problems outputs may use up a lot of hard-disk space, therefore full customization of the 
outputs is supported. 

This is controlled by the `output.yaml` config file found in the `config` folder of the workspace. This file contains
a full description of all customization options.

This PR enables customization of the `samples.csv` file, where all samples with a weight below a certain value can be automatically removed.

In many use cases, the `samples.csv` takes up the significant majority of the hard-disk space, which for large-scale
model-fitting problems can exceed gigabytes and be prohibitive to the analysis. This PR enables a workflow where this is not an issue.

### DIFF
--- a/autofit/non_linear/search/abstract_search.py
+++ b/autofit/non_linear/search/abstract_search.py
@@ -907,14 +907,13 @@ class NonLinearSearch(AbstractFactorOptimiser, ABC):
 
         samples = self.samples_from(model=model, search_internal=search_internal)
 
+
         try:
             instance = samples.max_log_likelihood()
         except exc.FitException:
             return samples
 
         if self.is_master:
-
-
 
             self.paths.save_samples(samples=samples)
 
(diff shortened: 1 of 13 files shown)


autofit/non_linear/search/abstract_search.py
```diff
@@ -913,7 +913,22 @@ def perform_update(
...
+            samples_weight_threshold = conf.instance["output"]["samples_weight_threshold"]
+
+            if samples_weight_threshold is not None:
+                samples_for_csv = samples_for_csv.samples_above_weight_threshold_from(
+                    weight_threshold=samples_weight_threshold
+                )
+
+                logger.info(
+                    f"Samples with weight less than {samples_weight_threshold} removed from samples.csv and not used to "
+                    f"compute parameter estimates errors, etc."
```
VascoSch92, line 928:
```suggestion
                    "compute parameter estimates errors, etc."

```
you don't need the f, as it is not an f-string :-)

rhayes777 (approved):

autofit/non_linear/search/abstract_search.py
Jammy2211: Thanks!

PR #971, opened 2024-03-19
feature/find groups

rhayes777:
Another attempt at concise that separates out the variables that cause branches of the model to differ


test_autofit/text/test_concise.py
```diff
@@ -17,6 +17,7 @@ def make_collection():
 
 
 def test_model_info(collection):
+    print(collection.info)
```
Jammy2211, line 20: :O

Jammy2211 (commented):
Looks like its pretty much there!

One issue is that tuple priors are being combined in a way that loses information (the 

```
            centre
                centre_0 - centre_1                                             GaussianPrior [547], mean = 0.0, sigma = 0.0001
```

These priors have two unique ids (presumably 547 and 548).

There is also a bit of a strange comibnation on fixed values:

```
            centre - ell_comps                                                  (0.0, 0.0)
```

rhayes777:
If we don't combine priors with different ids then concise model info will have essentially no effect.

Jammy2211:
> If we don't combine priors with different ids then concise model info will have essentially no effect.

The main point was to remove duplication of priors with the same id:

```
            light_profile_list
                0
                    centre
                        centre_0                                                UniformPrior [425], lower_limit = -0.1, upper_limit = 0.1
                        centre_1                                                UniformPrior [426], lower_limit = -0.1, upper_limit = 0.1
                    ell_comps
                        ell_comps_0                                             GaussianPrior [429], mean = 0.0, sigma = 0.3
                        ell_comps_1                                             GaussianPrior [430], mean = 0.0, sigma = 0.3
                    sigma                                                       0.01
                1
                    centre
                        centre_0                                                UniformPrior [425], lower_limit = -0.1, upper_limit = 0.1
                        centre_1                                                UniformPrior [426], lower_limit = -0.1, upper_limit = 0.1
                    ell_comps
                        ell_comps_0                                             GaussianPrior [429], mean = 0.0, sigma = 0.3
                        ell_comps_1                                             GaussianPrior [430], mean = 0.0, sigma = 0.3
                    sigma                                                       0.012294934136946248
                2
                    centre
                        centre_0                                                UniformPrior [425], lower_limit = -0.1, upper_limit = 0.1
                        centre_1                                                UniformPrior [426], lower_limit = -0.1, upper_limit = 0.1
                    ell_comps
                        ell_comps_0                                             GaussianPrior [429], mean = 0.0, sigma = 0.3
                        ell_comps_1                                             GaussianPrior [430], mean = 0.0, sigma = 0.3
                    sigma                                                       0.01511654054318462
                3
                    centre
                        centre_0                                                UniformPrior [425], lower_limit = -0.1, upper_limit = 0.1
                        centre_1                                                UniformPrior [426], lower_limit = -0.1, upper_limit = 0.1
                    ell_comps
                        ell_comps_0                                             GaussianPrior [429], mean = 0.0, sigma = 0.3
                        ell_comps_1                                             GaussianPrior [430], 
```

Which the PR does -- all these priors with ids 425, 29, etc are condensed into one.

But we dont want priors with unique ids to be compressed as I think this is useful info to a user.

rhayes777:
OK that's fine I can implement it that way

autofit/mapper/prior_model/representative.py
```diff
@@ -0,0 +1,69 @@
+from collections import defaultdict
+
+from ..prior.abstract import Prior
+
+
+def _value_hash(value) -> int:
+    """
+    When hashing values priors are hashed by their attributes, so that priors with the same attributes hash to the same
+    value. This is important for the grouping of priors in the model representation.
+    """
```
VascoSch92, line 10:
```suggestion
    """
    When hashing values, priors are hashed based on their attributes, so that priors with the same attributes hash to the same value. This is important for grouping priors in the model representation.
    """
```

autofit/mapper/prior_model/representative.py
```diff
@@ -0,0 +1,69 @@
...
+            )
+        )
+    return hash(value)
+
+
+def find_groups(path_value_tuples, limit=1):
+    """
+    Groups path-value tuples by their path, up to a limit. This is used to group priors in the model representation.
+
+    If multiple paths share the same suffix they are grouped together.
```
VascoSch92, line 24:
```suggestion
    If multiple paths share the same suffix, then they are grouped together.
```

autofit/mapper/prior_model/representative.py
```diff
@@ -0,0 +1,69 @@
...
+            tuple(
+                getattr(value, arg) for arg in value.__database_args__ if arg != "id_"
+            )
+        )
+    return hash(value)
+
+
+def find_groups(path_value_tuples, limit=1):
+    """
+    Groups path-value tuples by their path, up to a limit. This is used to group priors in the model representation.
```
VascoSch92, line 22:
```suggestion
    It groups path-value tuples by their path, up to a limit. This is used for grouping priors in the model representation.
```

autofit/mapper/prior_model/representative.py
```diff
@@ -0,0 +1,69 @@
...
+        path_value_tuples = _find_groups(path_value_tuples, position)
+    return path_value_tuples
+
+
+def _find_groups(path_value_tuples, position):
+    groups = defaultdict(list)
+    value_map = {}
+    paths = []
+    for path, value in path_value_tuples:
+        try:
```
VascoSch92, line 52: is there not a way to do it without try...expect ?

rhayes777: [It is easier to beg forgiveness than to ask permission](https://docs.python.org/3.5/glossary.html#term-eafp)

VascoSch92: haha ok 👌

rhayes777:
Done

Jammy2211 (approved):

PR #973, opened 2024-03-19
feature/sample type

### LINKED ISSUE #968: Wrong Samples type loaded via database
**Directory**

In the following script, we perform a fit using a nested sampler, but at the end of the script when we load the samples via the directory interface, we find the samples type is `SamplesPDF`:

https://github.com/Jammy2211/autofit_workspace_test/blob/main/bug/samples_directory_bug.py

The samples type should be `SamplesNest` -- but loading the wrong samples type certain methods are incorrect.

The reason for this is this function:

```
    @property
    def samples(self) -> SamplesPDF:
        """
        The samples of the search, parsed from a CSV containing individual samples
        and a JSON containing metadata.
        """
        if not self._samples:
            try:
                info_json = JSONOutput(
                    "info", self.files_path / "samples_info.json"
                ).dict

                with open(self.files_path / "samples.csv") as f:
                    sample_list = samples_from_iterator(csv.reader(f))

                self._samples = SamplesPDF.from_list_info_and_model(
                    sample_list=sample_list,
                    samples_info=info_json,
                    model=self.model,
                )
            except FileNotFoundError:
                raise AttributeError("No samples found")
        return self._samples
```

Therefore, we should update this function to return a samples type that corresponds to the correct type.

the samples above are loaded from `samples.csv`, which does not know the type of the search used to fit it and therefore cannot determine whether to load a `SamplesNest`, `SamplesMCMC` or something else.

I think there are two options for how we address this:

- Add the samples type to the `samples_info.json` file, which is loaded in the function above.
- Have the function above use the search (via `search.json`) to load the search and based on its instance choose the samples type.

**Database**

The following test script shows the same behaviour occurs when we load samples via a scraped .sqlite database:

https://github.com/Jammy2211/autofit_workspace_test/blob/main/bug/samples_scrape_bug.py

I don't think this occurs due to the same part of code shown above, but is instead a different module in the `database` package. Anyway, can you make it so we achieve the same behaviour of loading `SamplesNest` correctly.

rhayes777:
Resolves #968

### DIFF
--- a/autofit/database/model/common.py
+++ b/autofit/database/model/common.py
@@ -11,7 +11,7 @@ from .model import Object
 
 class Dict(Object):
     """
-    A dictionary
+    Represents a Python dictionary in the database
     """
 
     __tablename__ = "dict"
--- a/autofit/database/model/common.py
+++ b/autofit/database/model/common.py
@@ -1,0 +1,36 @@
+from typing import Union
+
+from ..sqlalchemy_ import sa
+
+from autofit.mapper.prior import abstract
+from autofit.mapper.prior_model import prior_model
+from autofit.mapper.prior_model import collection
+
+from .model import Object
+
+
+class Dict(Object):
+    """
+    A dictionary
+    """
+
+    __tablename__ = "dict"
+
+    id = sa.Column(
+        sa.Integer,
+        sa.ForeignKey("object.id"),
+        primary_key=True,
+        index=True,
+    )
+
+    __mapper_args__ = {"polymorphic_identity": "dict"}
+
+    def __call__(self):
+        return {child.name: child() for child in self.children}
+
+    @classmethod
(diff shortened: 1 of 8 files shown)


Jammy2211 (approved):

autofit/database/model/common.py
```diff
@@ -0,0 +1,36 @@
...
+from autofit.mapper.prior_model import prior_model
+from autofit.mapper.prior_model import collection
+
+from .model import Object
+
+
+class Dict(Object):
+    """
+    A dictionary
+    """
```
VascoSch92, line 15: The docstring is not so useful. Perhaps describe why this is different of a dictionary or why you define a dictionary :-)

rhayes777: Haha fair point. This was a copilot contribution

PR #981, opened 2024-03-22
Feature/samples summary in result

Jammy2211:
Make it so that all internal results loading use the `SamplesSummary` object instead of the `Samples` object wherever possible.

This has two benefits:

-Loading `samples.csv` /  `search_internal.dill` in order to load results when resuming a model-fit was very slow.

- Recent updates to `output.yaml` mean autofit can now run without outputting a `samples.csv` file or `search_internal.dill` file at all (to save hard disk space).  The samples summary file cannot be customized to not be output, meaning the information for resuming runs is always available.

This is motivated for search chaining prior passing, which by using the `SamplesSummary` resumes previous fits much faster.

### DIFF
--- a/autofit/non_linear/mock/mock_search.py
+++ b/autofit/non_linear/mock/mock_search.py
@@ -88,7 +88,7 @@ class MockSearch(NonLinearSearch):
                 # Return Chi squared
                 return -2 * log_likelihood
 
-        self.paths.save_samples_summary(self.samples_summary)
+        # self.paths.save_samples_summary(self.samples_summary)
 
         if self.save_for_aggregator:
             analysis.save_attributes(paths=self.paths)
--- a/autofit/non_linear/mock/mock_search.py
+++ b/autofit/non_linear/mock/mock_search.py
@@ -92,7 +92,7 @@ class MockSearch(NonLinearSearch):
                 # Return Chi squared
                 return -2 * log_likelihood
 
-        # self.paths.save_samples_summary(self.samples_summary)
+        self.paths.save_samples_summary(self.samples_summary)
 
         if self.save_for_aggregator:
             analysis.save_attributes(paths=self.paths)
@@ -104,8 +104,7 @@ class MockSearch(NonLinearSearch):
 
     def _fit(self, model, analysis):
         if self.fit_fast:
-            result = self._fit_fast(model=model, analysis=analysis)
-            return result
+            return self._fit_fast(model=model, analysis=analysis)
 
         if model.prior_count == 0:
             raise AssertionError("There are no priors associated with the model!")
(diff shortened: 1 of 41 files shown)


autofit/non_linear/samples/pdf.py
```diff
@@ -50,15 +50,19 @@ def __init__(
...
-            logging.warning(f"Could not create covariance matrix: {e}")
-            covariance_matrix = None
+
+        median_pdf_sample = Sample.from_lists(
+            model=self.model,
+            parameter_lists=[self.median_pdf(as_instance=False)],
+            log_likelihood_list=[self.max_log_likelihood_sample.log_likelihood],
+            log_prior_list=[self.max_log_likelihood_sample.log_prior],
+            weight_list=[self.max_log_likelihood_sample.weight],
+        )[0]
```
rhayes777, line 60: Seems silly to use a function called 'from_lists' only to take the first item

rhayes777:
I guess it makes a lot of sense for Result to take Paths; it starts to look a lot like the [SearchOutput](https://github.com/rhayes777/PyAutoFit/blob/main/autofit/aggregator/search_output.py) class used by the agregator.

What happens when a search has no name?

autofit/non_linear/mock/mock_search.py
```diff
@@ -1,167 +1,173 @@
...
+            except exc.FitException as e:
+                unit_vector[index] += 0.1
+                if unit_vector[index] >= 1:
+                    raise e
+                index = (index + 1) % model.prior_count
+
+        samples_summary = MockSamplesSummary(
+            #           sample_list=samples_with_log_likelihood_list(
+            #               self.sample_multiplier * fit, _make_samples(model)
+            #           ),
```
rhayes777, line 131: Commented code

autofit/non_linear/result.py
```diff
@@ -201,12 +260,63 @@ def dict(self) -> dict:
...
+
+        Returns
+        -------
+        The samples of the non-linear search.
+        """
+
+        # TODO : This needs to load a samples class based on the samples type.
+
+        if self._samples is not None:
+            return self._samples
```
rhayes777, line 287: Could use cached_property

autofit/non_linear/result.py
```diff
@@ -201,12 +260,63 @@ def dict(self) -> dict:
...
+
+        try:
+            Samples.from_csv(
+                paths=self.paths,
+                model=self.model,
+            )
+        except FileNotFoundError:
+            pass
```
rhayes777, line 295: Should the return type be Optional?

rhayes777 (approved):

autofit/non_linear/result.py
Jammy2211: Not quite, this is because samples can be passed into `__init__` meaning we dont load via csv (e.g. to avoid I/O)

rhayes777: Ah ofc

PR #986, opened 2024-03-26
Feature/keep the noise down

Jammy2211:
Adds warnings and code tweaks to avoid noise on the command line when running unit tests and the code.


rhayes777 (approved):
I basically can't tell what's changed because of the other PRs in this one. If possible create each new PR by branching from main

PR #989, opened 2024-03-29
Feature/result cast

Jammy2211:
New API for how the `Analysis` class returns a `Return`.

Now, the `Analysis` class has a `Result` class attribute, which a user can overwrite with their own custom `Result` class. 

This removes a number of unecessary `make_result` functions that end up populating `Analysis` objects and is a cleaner API for users, as shown in the docs updated in this PR.


autofit/non_linear/result.py
```diff
@@ -163,50 +198,141 @@ def model_bounded(self, b: float) -> AbstractPriorModel:
...
+        if self._samples is not None:
+            return self._samples
+
+        try:
+            Samples.from_csv(
+                paths=self.paths,
+                model=self.model,
+            )
+        except FileNotFoundError:
+            pass
```
rhayes777, line 311: Did you mean to return this?

test_autofit/interpolator/test_covariance.py
```diff
@@ -31,89 +31,89 @@ def test_covariance_matrix(interpolator):
...
+#         interpolator.covariance_matrix(), interpolator.inverse_covariance_matrix()
+#     )
+#     print(identity)
+#     assert np.allclose(
+#         identity,
+#         np.eye(9),
+#     )
+
+
+# def test_interpolate(interpolator):
```
rhayes777, line 45: Ooof why is this all commented out?

rhayes777 (approved):
We need to fix those interpolator tests

autofit/example/analysis.py
```diff
@@ -43,6 +53,34 @@ def log_likelihood_function(self, instance: af.ModelInstance) -> float:
         -------
         The log likelihood value indicating how well this model fit the dataset.
         """
+        model_data_1d = self.model_data_1d_from(instance=instance)
+
+        residual_map = self.data - model_data_1d
+        chi_squared_map = (residual_map / self.noise_map) ** 2.0
+        log_likelihood = -0.5 * sum(chi_squared_map)
```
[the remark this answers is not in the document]
Jammy2211: This is for the example module, which users might read, so intentionally simple.

PR #1021, opened 2024-07-01
feature/prior array

rhayes777:
Adds the Array class which functions as a PriorModel that creates numpy arrays of floats.

An array is defined by its shape and a prior.

```python
array = af.Array(
    shape=(2, 2),
    prior=af.GaussianPrior(mean=0.0, sigma=1.0),
)
```

The prior is copied to each index meaning that a 2x2 array has four independent priors.

```python
print(array.prior_count)
> 4
```

Arrays can be accessed and modified using indexing.

```python
array[0, 0] = 1.0
print(array.prior_count)
> 3
```

They can be instantiated just as any other model class
```
instance = array.instance_from_prior_medians()
print(instance)
> [
     [1.0, 0.0],
     [0.0, 0.0],
  ]

### CODE REVIEW

**Review by Jammy2211 — APPROVED** [review]
Don't merge yet -- I'll build a test case this week and see whether any issues come up.

Couple of things to think about, albeit best not to worry about them until we've done some testing:

1) Does this all play nicely with our JAX / Pytrees wrapping?
2) How does this play with search chaining API?
3) Is there a clean interface with the default-prior yaml configs?

**Comment by rhayes777** [comment]
Need to check JAX and search chaining. We mostly do config on the type -> prior relationship so it was hard to see how to carry this over

**Comment by Jammy2211** [comment]
I would make it so that the config default gives the same prior to everything on the numpy array.

We can worry about JAX and search chaining later on then!

**Comment by rhayes777** [comment]
> I would make it so that the config default gives the same prior to everything on the numpy array.
> 
> We can worry about JAX and search chaining later on then!

So at the moment you define one prior and that gets copied to every slot. Would we want a config to determine the default for all Arrays globally?

**Comment by rhayes777** [comment]
Should now work with prior passing and JAX

File: autofit/mapper/prior_model/array.py [hunk 1]
```diff
@@ -0,0 +1,215 @@
...
+        prior: Optional[Prior] = None,
+    ):
+        """
+        An array of priors.
+
+        Parameters
+        ----------
+        shape : (int, int)
+            The shape of the array.
+        prior : Prior
```
VascoSch92, line 25: can you describe what happens if the parameter prior is not provided?

autofit/mapper/prior_model/array.py
```diff
@@ -0,0 +1,215 @@
...
+            The prior of every entry in the array.
+        """
+        super().__init__()
+        self.shape = shape
+        self.indices = list(np.ndindex(*shape))
```
VascoSch92, line 30: you are converting `self.indices` to a list but you are usint the typing `Tuple[int, ...]` everywhere.

PR #1034, opened 2024-08-05
naive implementation of physical_centres_from

rhayes777:
Pretty sure this does what you want as it uses the model passed to the samples for each grid search run. For any prior which is a grid prior this new function would give you the mean of that prior for each model

### DIFF
--- a/autofit/non_linear/grid/grid_search/result.py
+++ b/autofit/non_linear/grid/grid_search/result.py
@@ -41,6 +41,25 @@ class GridSearchResult:
 
         self.parent = parent
 
+    @as_grid_list
+    def physical_centres_lists_from(self, path: str):
+        """
+        Get the physical centres of the grid search from a path to an attribute of the instance in the samples.
+
+        Parameters
+        ----------
+        path
+            The path to the attribute to get from the instance
+
+        Returns
+        -------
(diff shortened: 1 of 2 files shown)


rhayes777:
Obviously requires more testing and generalisation to support tuple input; I just wanted to check I've understood correctly

Jammy2211:
I have tested the implementation by using the following code in the relevent part of the PyAutoLens source code:

```
    @property
    def y(self) -> List[float]:
        """
        The y coordinates of the physical values of the subhalo grid, where each value is the centre of a grid cell.

        These are the `centre` coordinates of the dark matter subhalo priors.
        """
        return self.physical_centres_lists_from(
            path="galaxies.subhalo.mass.centre.centre_0"
        )

    @property
    def x(self) -> List[float]:
        """
        The x coordinates of the physical values of the subhalo grid, where each value is the centre of a grid cell.

        These are the `centre` coordinates of the dark matter subhalo priors.
        """
        return self.physical_centres_lists_from(
            path="galaxies.subhalo.mass.centre.centre_1"
        )
```

This PR gives the desired behaviour, in that it recovers the values for y and x correctly and can be used for other quantities in the model.

Jammy2211 (approved):

PR #1043, opened 2024-09-06
Feature/build fixes

Jammy2211:
Various fixes to make build server runs pass.

### DIFF
--- a/autofit/non_linear/result.py
+++ b/autofit/non_linear/result.py
@@ -311,7 +311,7 @@ class Result(AbstractResult):
 
         try:
             return self.paths.load_search_internal()
-        except FileNotFoundError:
+        except (FileNotFoundError, RecursionError):
             pass
 
     @property
--- a/autofit/non_linear/result.py
+++ b/autofit/non_linear/result.py
@@ -311,7 +311,7 @@ class Result(AbstractResult):
 
         try:
             return self.paths.load_search_internal()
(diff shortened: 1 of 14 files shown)


autofit/non_linear/result.py
```diff
@@ -311,7 +311,7 @@ def search_internal(self):
 
         try:
             return self.paths.load_search_internal()
-        except FileNotFoundError:
+        except (FileNotFoundError, RecursionError):
```
rhayes777, line 314: Oh yuck why is a RecursionError occurring? We should get to the bottom of that really

autofit/non_linear/search/mcmc/zeus/search.py
```diff
@@ -274,13 +275,23 @@ def samples_via_internal_from(self, model, search_internal=None):
 
         search_internal = search_internal or self.paths.load_search_internal()
 
-        auto_correlations = self.auto_correlations_from(search_internal=search_internal)
+        if os.environ.get("PYAUTOFIT_TEST_MODE") == "1":
```
rhayes777, line 278: I feel like we should make this a function

rhayes777 (approved):

autofit/non_linear/result.py
Jammy2211: I will remove it and then merge, and post an issue when it crops up again.

PR #1063, opened 2024-10-11
feature/free parameter override

### LINKED ISSUE #1015: Customizing Priors After analysis.with_free_parameter
We often use the following API to extend a model over summed analysis classes:

```
...

analysis = sum(analysis_list)

dataset_model = af.Model(al.DatasetModel)

model = af.Collection(
    dataset_model=dataset_model,
    galaxies=af.Collection(lens=lens, source=source)
)

analysis = analysis.with_free_parameters(
    model.dataset_model.grid_offset
)
```

I now have a use case where I want to customize the prior on the `grid_offset` of the `DatasetModel` corresponding to the second `analysis` only. 

I don't think we currently have an API that allows us to do this?

I cannot achieve the desired behaviour by editing prior configs, because I want the prior on one `grid_offset` to differ from the other.

Let me know if we have an API hidden, else lets think about how we extend the API.

rhayes777:
Resolves #1015

Specific model or parameter overrides can now be defined by calling get-set on the free parameter analysis.

```python
analysis = sum(analysis_list)

dataset_model = af.Model(al.DatasetModel)

model = af.Collection(
    dataset_model=dataset_model,
    galaxies=af.Collection(lens=lens, source=source)
)

analysis = analysis.with_free_parameters(
    model.dataset_model.grid_offset
)

analysis[0][model.dataset_model.background_sky_level] = 1.0

### DIFF
--- a/autofit/mapper/prior_model/abstract.py
+++ b/autofit/mapper/prior_model/abstract.py
@@ -1523,6 +1523,14 @@ class AbstractPriorModel(AbstractModel):
         """
         return sorted(priors, key=lambda prior: self.path_for_prior(prior))
 
+    def path_for_object(self, obj) -> Optional[Path]:
+        for path, instance in self.path_instance_tuples_for_class(
+            object, ignore_children=False
+        ):
+            if instance is obj:
+                return path
+        return None
+
     def path_for_prior(self, prior: Prior) -> Optional[Path]:
         """
         Find a path that points at the given tuple.
--- a/autofit/mapper/prior_model/abstract.py
+++ b/autofit/mapper/prior_model/abstract.py
@@ -1524,6 +1524,18 @@ class AbstractPriorModel(AbstractModel):
         return sorted(priors, key=lambda prior: self.path_for_prior(prior))
 
     def path_for_object(self, obj) -> Optional[Path]:
+        """
+        Find a path that points at the given object.
+
+        Parameters
+        ----------
+        obj
+            An object in the model.
+
+        Returns
+        -------
+        A path, a series of attributes that point to one location of the object.
+        """
         for path, instance in self.path_instance_tuples_for_class(
             object, ignore_children=False
         ):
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by Jammy2211** [comment]
There is a bug in this updated script:

https://github.com/Jammy2211/autolens_workspace/blob/main/scripts/advanced/multi/modeling/features/dataset_offsets.py

```
dataset_model = af.Model(al.DatasetModel)

model = af.Collection(
    dataset_model=dataset_model,
    galaxies=af.Collection(lens=lens, source=source)
)

"""
We now make the dataset offsets vary across every analysis object.
"""
analysis = analysis.with_free_parameters(model.dataset_model.grid_offset)

"""
The default prior on a `DatasetModel`'s offsets are actually not a prior, but fixed values of (0.0, 0.0).

We must therefore manually overwrite the prior on the offsets of datasets we wish to be included. 

We updated the priors on the offsets of the second dataset to be GaussianPrior's with mean 0.0" and sigma 0.1".
"""
analysis[1][model.dataset_model.grid_offset.grid_offset_0] = af.UniformPrior(
    lower_limit=-0.1, upper_limit=0.1
)
analysis[1][model.dataset_model.grid_offset.grid_offset_1] = af.UniformPrior(
    lower_limit=-0.1, upper_limit=0.1
)
```

Basically, if I run the code above, the prior on `model.dataset_model.grid_offset.grid_offset_1` is not updated, but the one on `model.dataset_model.grid_offset.grid_offset_0` is.

However, if I comment out the first prior:

```
#analysis[1][model.dataset_model.grid_offset.grid_offset_0] = af.UniformPrior(
 #   lower_limit=-0.1, upper_limit=0.1
#)
analysis[1][model.dataset_model.grid_offset.grid_offset_1] = af.UniformPrior(
    lower_limit=-0.1, upper_limit=0.1
)
```

The prior on `model.dataset_model.grid_offset.grid_offset_1` is set correctly.

It seems that setting a second prior over writes or resets a first prior? Maybe this is specific to tuples as well?

Jammy2211 (approved):

Issue #1087, opened 2025-01-15
Aggregator for csv building


Jammy2211:
You should first read this issue the `AggregatorPng`, as the general design and idea is the same:

https://github.com/rhayes777/PyAutoFit/issues/1086

**Use Case:**

Similar to .png splicing, it was common for me to want to view the numerical results of lens modeling in a single
.csv file, rather than navigating the `output` folder to find the information I needed.

**Implementation:**

Here is an example of the .csv file for the 2 example images on the `agg_png_csv` URL above:

https://github.com/Jammy2211/autolens_workspace_test/blob/main/agg_png_csv/result.csv

This .csv file contains headers which come from different parts of the `dataset` and `output` folders, thus navigating
the folders to find the information requires combining information from multiple files.

Here is the example .csv maker python file I was using, which is a bit of a mess, but works:

https://github.com/Jammy2211/autolens_workspace_test/blob/main/agg_png_csv/csv_make.py

**AggregatorCSV**

I am picturing an `AggregatorCSV` class that would take the `output` folder and help build the .csv file.

**Using Output Folder**

For the example on the workspace, all information used to build the .csv file comes from the `info.json` and
`result.json` files in the `dataset` folder.

So you would just need to write an `AggregatorCSV` that navigates the `dataset` folder and loads the `info.json` and
`result.json` files, and produces the .csv file.

My .csv building never used the `output` folder, but this was because I wrote a quite complicated pipeline (in a hurry)
which output all results to the `dataset` folder.

I think the general use case would be that the `output` folder is used to extract all information whenever possible.

For example, the `einstein_radius_max_lh` is stored here:

![image](https://github.com/user-attachments/assets/382e6775-4ddc-401b-ba1b-6db1701513ef)

Where `samples_summary.json` has all the information on the model and instance, with an einstein radius at this part:

```
    "median_pdf_sample": {
            "type": "instance",
            "class_path": "autofit.non_linear.samples.sample.Sample",
            "arguments": {
                "log_likelihood": 21775.911731379132,
                "log_prior": 1.4651345875102018,
                "weight": 6.53867998265804e-05,
                "kwargs": {
                    "type": "dict",
                    "arguments": {
                        "galaxies.lens.bulge.profile_list.59.centre.centre_0": -0.07025111305343998,
                        "galaxies.lens.bulge.profile_list.59.centre.centre_1": -0.02258690706633907,
[...]

PR #1089, opened 2025-01-17
feature/aggregate summary

rhayes777:
Introduces a summary CSV over chosen fields from search output using the aggregator as per #1087

Does not currently support numerical errors as the current output is not sufficient and changing the output results in failing tests.

Usage:

```python
aggregator = Aggregator.from_directory(directory)
summary = AggregateSummary(aggregator)

# Simply add the median pdf value by its path
summary.add_column("galaxies.lens.bulge.centre.centre_0")

# Add a median pdf value but give it a specific name
summary.add_column(
    "galaxies.lens.bulge.centre.centre_1",
    name="centre_1",
)

# If a latent_summary.json file is present latent values can be used implicitly
summary.add_column(
    "latent.value",
)

# Computed column values can also be added
def compute(samples):
     return 1
summary.add_computed_column(
    "computed",
    compute,
)

# Save the CSV
summary.save(output_path)

### DIFF
--- a/autofit/aggregator/aggregate_csv.py
+++ b/autofit/aggregator/aggregate_csv.py
@@ -116,7 +116,7 @@ class Row:
         return row
 
 
-class AggregateSummary:
+class AggregateCSV:
     def __init__(self, aggregator: Aggregator):
         """
         Summarise results from the aggregator as a CSV.
--- a/autofit/aggregator/aggregate_summary.py
+++ b/autofit/aggregator/aggregate_summary.py
@@ -8,31 +8,43 @@ import csv
 class Column:
     def __init__(
         self,
-        column: str,
+        argument: str,
         name: Optional[str] = None,
     ):
-        self.column = column
-        self.name = name or column.replace(
+        self.argument = argument
+        self.name = name or argument.replace(
             ".",
             "_",
         )
 
+    @property
+    def path(self):
+        return tuple(self.argument.split("."))
+
 
 class AggregateSummary:
     def __init__(self, aggregator: Aggregator):
-        self.aggregator = aggregator
+        self._aggregator = aggregator
+        self._columns = []
 
     def add_column(self, argument):
-        pass
+        self._columns.append(Column(argument))
+
+    @property
+    def fieldnames(self):
+        return ["id"] + [column.name for column in self._columns]
 
     def save(self, path):
         with open(path, "w") as f:
             writer = csv.DictWriter(
                 f,
-                fieldnames=["id"],
+                fieldnames=self.fieldnames,
             )
             writer.writeheader()
-            for result in self.aggregator:
+            for result in self._aggregator:
                 writer.writerow({"id": result.id})
                 samples_summary = result.value("samples_summary")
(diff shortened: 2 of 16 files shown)

### CODE REVIEW

**Review by Jammy2211 — APPROVED** [review]
I would maybe suggest `AggregateCSV` rather than `AggregateSummary` as the latter seems generic, assuming that we'll also have an `AggregatePNG` soon too.

Unless you think the API you've chosen makes more sense for some reason? I

**Comment by rhayes777** [comment]
Yeah makes sense

**Comment by Jammy2211** [comment]
I began writing an example script using this for PyAutoGalaxy.

First, I wrote my own `Subplot` object in PyAutoGalaxy, specific to its `subplot_fit.png`:

```
class SubplotFit(Enum):
    """
    The subplots that can be extracted from the subplot_fit image.

    The values correspond to the position of the subplot in the 4x3 grid.
    """

    Data = (0, 0)
    SignalToNoiseMap = (1, 0)
    ModelImage = (2, 0)
    NormalizedResidualMap = (0, 1)
    NormalizedResidualMapOneSigma = (1, 1)
    ChiSquaredMap = (2, 1)
```

However, the `SubplotFitImage` in autofit also had coded into it the subplot shape:

```
class SubplotFitImage:
    def __init__(self, image: Image.Image):
        """
        The subplot_fit image associated with one fit.

        Parameters
        ----------
        image
            The subplot_fit image.
        """
        self._image = image

        self._single_image_width = self._image.width // 4
        self._single_image_height = self._image.height // 3
```

The API for how a user customizes the png splicing makes a lot of sense:

```
image = agg_png.extract_image(
    ag.agg.SubplotFit.Data,
    ag.agg.SubplotFit.ModelData,
    ag.agg.SubplotFit.NormalizedResidualMap,
)
```

But the `SubplotFitImage` is not updated to account for the fact its using autogalaxy.

Is it feasible to make it so `SubplotFit(Enum)` contains all information we need to set up this interface, so in PyAutoGalaxy my method would read:

```
class SubplotFit(Enum):
    """
    The subplots that can be extracted from the subplot_fit image.

    The values correspond to the position of the subplot in the 4x3 grid.
    """

    filename = "subplot_fit.png"

    subplot_shape = (3, 2)

    Data = (0, 0)
    SignalToNoiseMap = (1, 0)
    ModelImage = (2, 0)
    NormalizedResidualMap = (0, 1)
    NormalizedResidualMapOneSigma = (1, 1)
    ChiSquaredMap = (2, 1)
```

This would mean that `SubplotFitImage` in autofit, which looks like it contains general functionality to do the splicing, would need a `subplot_shape` input.

**Comment by rhayes777** [comment]
I was thinking about this the other day and maybe the way to handle it is to have a mapping in each place the enum is used?

Something like
```python
class Subplot(Enum):
    Data = 1
    SignalToNoiseMap = 2
    ...


position_map = {
    Subplot.Data: (0, 0),
    SignalToNoiseMap: (1, 0),
    ...
}
```

Alternatively we could bake everything in to the enum values. You can even use class instances to provide the values giving a bit more structure and flexibility.

Issue #1086, opened 2025-01-15
Aggregator for png splicing


Jammy2211:
**Use Case:**

To model 500 strong lens from Euclid, I found the workflow of navigating the `output` folder to inspect results
slow,

Furthermore, it was common for the information I needed to be spread across multiple images, leading to more clicks:

![image](https://github.com/user-attachments/assets/7fc46003-cbb9-4853-8fb8-25be4a4222a7)

I developed a workflow which went to .png files I had output in the pipeline, extracted cut-outs showing the parts of the
fits of each individual lens I was interested in and saved them to .png files in a single folder that I could inspect
quickly.

The following URL shows an example of how

https://github.com/Jammy2211/autolens_workspace_test/tree/main/agg_png_csv

If you run the file `png_to_single_line.py`, you'll make a folder `png` containing images of the two example images
that enable quick inspection:

![image](https://github.com/user-attachments/assets/7d8a6ba7-602c-45e3-b838-a418b4b26972)

This uses the PIL library to open .png subplots showing many images and extract the individual images of each subplot
and save them to a new folder with a new figure and file name.

The images used by the code above to extract subplots are in the `dataset` folder, because for Euclid modeling
this is where I chose to store the images I used to extract subplots. However, I think a better implementation
would revolve entirely around the standard `output` folder produced by **PyAutoFit**.

**AggregatorPng**

I am picturing an `AggregatorPng` class that mirrors the API of the `Aggregator` class, which will allow a user
to quickly extract images from the image subplots inside the `output` folder and save them to a new folder.

The GitHub above has an `output` folder which contains the example outputs for the two lens, which is what the
`AggregatorPng` class can be developed to work with.

**Example Query 1:**

The first query would be to produce the following .png showing the fit to a given lens:

![image](https://github.com/user-attachments/assets/615de9e4-29d0-4377-bae1-8f6c85180cfd)

All of the information needed to make this plot is contained in a single .png file of each lens fit, called
`subplot_fit.png`. For this image, the source pixelization is shown, so we need the results of `source_pix[2]`:

![image](https://github.com/user-attachments/assets/d2ac08b7-153e-4d11-8d0a-42128f5a1a0e)

So the first query for `AggregatorPng` would be something like:

```
agg = AggregatorPng.from_directory(
    directory=path.join("output),
)

image = agg.extract_images(
    image_type="subplot_fit",
    new_order = [5, 6, 11], # These are actually correct and I have chosen them based on subplot_fit.png
[...]

PR #1090, opened 2025-01-17
feature/aggregate images

rhayes777:
Aggregator png splicing as per #1086

Does not yet support customising output tiling.

Usage:

```python
from autofit.aggregator.aggregate_images import AggregateImages, Subplot

aggregator = Aggregator.from_directory(directory)
aggregate = AggregateImages(aggregator)

image = aggregate.extract_image(
    Subplot.NormalizedResidualMap,
    Subplot.SourcePlaneNoZoom,
    Subplot.SourceModelImage,
)

image.save("combined.png")

### CODE REVIEW

**Comment by Jammy2211** [comment]
This is mostly good, my main concern is its making a new API for images that does not mirror the existing one.

Currently, users are used to seeing the image `subplot_fit.png`.

They are used to making this plot via Python code like:

```
fit_plotter = aplt.FitImagingPlotter(fit=fit)

fit_plotter.subplot_fit()
```

They are also used to specifying individual plots via:

```
fit_plotter = aplt.FitImagingPlotter(fit=fit)

fit_plotter.figures_2d(
    data=True,
    noise_map=True,
    signal_to_noise_map=True,
    model_image=True,
    residual_map=True,
    normalized_residual_map=True,
    chi_squared_map=True,
)
``` 

Currently, it feels like `Subplot` and commands like `Subplot.NormalizedResidualMap` are duplicating the API.

However, the APIs cant be too tightly coupled as certain images use function calls like this:

```
fit_plotter.figures_2d_of_planes(plane_index=0, subtracted_image=True)
```

Maybe lower casing the `Subplot` variables for now would get us close to a mirrored API:

```
image = aggregate.extract_image(
    Subplot.normalized_residual_map,
)
```

I'll give this a bit more thought, I could also imagine it being some sort of dictionary where the key is `subplot_fit.png` and it has the mappings somehow... idk

**Comment by rhayes777** [comment]
Easy enough to lowercase for consistency.

tbh I think the more natural way to implement the original visualisation API would be using enums. It's much neater than an ever growing list of default False bools

**Comment by Jammy2211** [comment]
What does the enums API look like?

**Comment by rhayes777** [comment]
```python
image = aggregate.extract_image(
    Subplot.NormalizedResidualMap,
    Subplot.SourcePlaneNoZoom,
    Subplot.SourceModelImage,
)
```

Subplot is an Enum

Jammy2211 (approved):
Check unit tests before merge, failing locally for me.

PR #1096, opened 2025-01-24
feature/search children

### LINKED ISSUE #1093: Generalize Aggregator directory and scrape API
Two examples of database use on the `autolens_workspace` only work when a .sqlite database is built and used:

https://github.com/Jammy2211/autolens_workspace/blob/release/scripts/results/examples/data_fitting.py
https://github.com/Jammy2211/autolens_workspace/blob/release/scripts/results/examples/models.py

This is because they use custom objects, like `FitImagingAgg`, when streamline standard calculations using the aggregator for the user:

https://github.com/Jammy2211/PyAutoLens/blob/main/autolens/aggregator/fit_imaging.py

I think the core part of the API where support for sqlite-only is here:

https://github.com/Jammy2211/PyAutoGalaxy/blob/main/autogalaxy/aggregator/agg_util.py

```
    fit_list = [fit] if not fit.children else fit.children

    if fit.value(name="adapt_images.adapt_images") is None:
        return [None] * len(fit_list)

    adapt_images_list = []

    for fit in fit_list:
```

I assume something here is tied to how the sqlite stores results.

Could you make it so that the above example scripts, and therefore these custom `Agg` objects, support Aggregator use which just reads directories? 

Its common to have users with ~5 - 10 datasets where the sqlite database is just adding complexity and over head without much benefit.

To generate the results for these example scripts you will need to run this:

https://github.com/Jammy2211/autolens_workspace/blob/release/scripts/results/start_here.py

rhayes777:
Should resolve #1093

SearchOutput (i.e. from a directory) and Fit (i.e. from the database) had different names for this attribute/property

I've made the name consistent (children) but also added a common ABC which asserts some interface that should be shared. I've swapped Fit (from the database) for Fit (the new ABC) in the autofit init such that the new type should be used implicitly.

### DIFF
--- a/autofit/__init__.py
+++ b/autofit/__init__.py
@@ -25,9 +25,8 @@ from .non_linear.samples import Sample
 from .non_linear.samples import load_from_table
 from .non_linear.samples import SamplesStored
 from .database.aggregator import Aggregator
-from .database.model import Fit
 from .database.aggregator import Query
-from .database.model.fit import Fit
+from autofit.aggregator.fit_interface import Fit
 from .aggregator.search_output import SearchOutput
 from .mapper import prior
 from .mapper.model import AbstractModel
--- a/autofit/aggregator/base.py
+++ b/autofit/aggregator/base.py
@@ -156,7 +156,7 @@ class AggBase(ABC):
         """
 
         def func_gen(fit: af.Fit, total_samples: int) -> List[object]:
-            samples = fit.value(name="samples")
+            samples = fit.samples
 
             return [
                 self.object_via_gen_from(
--- a/autofit/aggregator/base.py
+++ b/autofit/aggregator/base.py
@@ -80,7 +80,7 @@ class AggBase(ABC):
         """
 
         def func_gen(fit: af.Fit, minimum_weight: float) -> List[object]:
-            samples = fit.value(name="samples")
+            samples = fit.samples
 
             weight_list = []
 
@@ -117,7 +117,7 @@ class AggBase(ABC):
         """
 
         def func_gen(fit: af.Fit, minimum_weight: float) -> List[object]:
-            samples = fit.value(name="samples")
+            samples = fit.samples
 
             all_above_weight_list = []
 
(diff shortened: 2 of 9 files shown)


rhayes777:
Oh yeah I think I fixed a similar issue to get the other script working

rhayes777:
This was fixed by directly calling the 'samples' attribute

Jammy2211:
So it looks like it worked for the `data_fitting.py` example, but when I run it on the other script that needs this change, `models.py` I get the same error?

https://github.com/Jammy2211/autolens_workspace/blob/release/scripts/results/examples/models.py

```
The above exception was the direct cause of the following exception:

  File "/mnt/c/Users/Jammy/Code/PyAuto/autolens_workspace/scripts/results/examples/models.py", line 162, in <module>
    for tracer_gen, weight_gen in zip(tracer_list_gen, weight_list_gen):
  File "/mnt/c/Users/Jammy/Code/PyAuto/PyAutoFit/autofit/aggregator/base.py", line 120, in func_gen
    samples = fit.value(name="samples")
  File "/mnt/c/Users/Jammy/Code/PyAuto/PyAutoFit/autofit/aggregator/search_output.py", line 159, in value
    return item.value
  File "/mnt/c/Users/Jammy/Code/PyAuto/PyAutoFit/autofit/aggregator/file_output.py", line 59, in value
    return np.loadtxt(self.path, delimiter=",")
  File "/home/jammy/venvs/PyAuto/lib/python3.10/site-packages/numpy/lib/npyio.py", line 1373, in loadtxt
    arr = _read(fname, dtype=dtype, comment=comment, delimiter=delimiter,
  File "/home/jammy/venvs/PyAuto/lib/python3.10/site-packages/numpy/lib/npyio.py", line 1016, in _read
    arr = _load_from_filelike(
ValueError: could not convert string 'galaxies.lens.mass.centre.centre_0' to float64 at row 0, column 1.
```

So this like of the stack isn't in the exception above:

```
  File "/mnt/c/Users/Jammy/Code/PyAuto/PyAutoFit/autofit/aggregator/base.py", line 120, in func_gen
    samples = fit.value(name="samples")
```

So maybe theres another line of autofit that needs updating?

Jammy2211 (approved):

PR #1105, opened 2025-02-21
feature/aggregate image

### LINKED ISSUE #1086: Aggregator for png splicing

rhayes777:
Resolves #1086

You can now:

- Output to a folder
```python
aggregate.output_to_folder(
    output_directory,
    Subplot.Data,
    Subplot.SourcePlaneZoomed,
    Subplot.SourceModelImage,
)
```

- Optionally specify the name to use:
```python
aggregate.output_to_folder(
    output_directory,
    Subplot.Data,
    Subplot.SourcePlaneZoomed,
    Subplot.SourceModelImage,
    name="id",
)
```

- Output custom images by passing a list of images or a function that takes a search output:
```python
image = Image.new("RGB", (10, 10), "white")
images = [image for _ in aggregator]

result = aggregate.extract_image(
    Subplot.Data,
    Subplot.SourcePlaneZoomed,
    Subplot.SourceModelImage,
    images,
)
```

```python
image = Image.new("RGB", (10, 10), "white")
images = [image for _ in aggregator]

result = aggregate.extract_image(
    Subplot.Data,
    Subplot.SourcePlaneZoomed,
    Subplot.SourceModelImage,
    images,
)
```

### DIFF
--- a/autofit/aggregator/aggregate_images.py
+++ b/autofit/aggregator/aggregate_images.py
@@ -1,3 +1,4 @@
+import re
 import sys
 from enum import Enum
 from typing import Optional, List, Union, Callable, Type
@@ -9,7 +10,7 @@ from autofit.aggregator.search_output import SearchOutput
 from autofit.aggregator.aggregator import Aggregator
 
 
-class Subplot(Enum):
+class SubplotFit(Enum):
     """
     The subplots that can be extracted from the subplot_fit image.
 
@@ -34,7 +35,7 @@ class SubplotFitImage:
     def __init__(
         self,
         image: Image.Image,
-        suplot_type: Type[Subplot] = Subplot,
+        suplot_type: Type[SubplotFit] = SubplotFit,
     ):
         """
         The subplot_fit image associated with one fit.
@@ -100,10 +101,11 @@ class AggregateImages:
             The aggregator containing the fit results.
         """
         self._aggregator = aggregator
+        self._source_images = None
 
     def extract_image(
         self,
-        *subplots: Union[Subplot, List[Image.Image], Callable],
+        *subplots: Union[Enum, List[Image.Image], Callable],
         subplot_width: Optional[int] = sys.maxsize,
     ) -> Image.Image:
         """
@@ -145,7 +147,7 @@ class AggregateImages:
     def output_to_folder(
         self,
         folder: Path,
-        *subplots: Union[Subplot, List[Image.Image], Callable],
+        *subplots: Union[SubplotFit, List[Image.Image], Callable],
         subplot_width: Optional[int] = sys.maxsize,
         name: str = "name",
     ):
@@ -187,7 +189,7 @@ class AggregateImages:
     def _matrix_for_result(
         i: int,
         result: SearchOutput,
-        *subplots: Union[Subplot, List[Image.Image], Callable],
+        *subplots: Union[SubplotFit, List[Image.Image], Callable],
         subplot_width: int = sys.maxsize,
     ) -> List[List[Image.Image]]:
         """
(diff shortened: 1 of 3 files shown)


autofit/aggregator/aggregate_images.py
```diff
@@ -38,8 +42,8 @@ def __init__(self, image: Image.Image):
         """
         self._image = image
 
-        self._single_image_width = self._image.width // 3
-        self._single_image_height = self._image.height // 2
+        self._single_image_width = self._image.width // 4
+        self._single_image_height = self._image.height // 3
```
rhayes777, line 47: I'm not sure why you changed this but I might have undone something here? The values you set made it inconsistent with the subplot indices defined in the Subplot Enum values

Jammy2211: So in my autogalaxy example I have two Subplot objects now:

```
from enum import Enum

class SubplotDataset(Enum):
    """
    The subplots that can be extracted from the subplot_fit image.

    The values correspond to the position of the subplot in the 4x3 grid.
    """

    data = (0, 0)
    data_log_10 = (1, 0)
    noise_map = (2, 0)
    psf = (0, 1)
    psf_log_10 = (1, 1)
    signal_to_noise_map = (2, 1)
    over_sample_size_lp = (0, 2)
    over_sample_Size_pixelization = (1, 2)

class SubplotFit(Enum):
    """
    The subplots that can be extracted from the subplot_fit image.

    The values correspond to the position of the subplot in the 4x3 grid.
    """

    data = (0, 0)
    signal_to_noise_map = (1, 0)
    model_image = (2, 0)
    normalized_residual_map = (0, 1)
    normalized_residual_map_one_sigma = (1, 1)
    chi_squared_map = (2, 1)
```

One has shape (3,3) the other (3, 2), and in general other subplots could be different shapes.

I think I updated them cause I was using `SubplotFit`, but do we basically need these values to come from the `Subplot` object itself? The values of 3 and 2 above I assume are hard-coded in but need to change for each subplot?

rhayes777: Do these two have separate output filenames? I guess we might also need to dynamically change the file we load from depending on the type of the enum?

Jammy2211 (approved):

PR #1108, opened 2025-02-25
Feature/sensitivity batch

Jammy2211:
Allows for sensitivity mapping to have integer batch values, so that it can easily be spread out over multipole nodes of a super computer.


rhayes777 (approved):
I wonder if there is a more elegant way to do this where we specify some job handler that decides how to distribute jobs?

Jammy2211:
I think there will be, at the moment i'm trying to get something that works and test end-to-end, and then reassess the implementation.

PR #1124, opened 2025-03-28
Feature/checkpoint deleting

Jammy2211:
This PR removes legacy code that manually deletes checkpoint files for Nautilus and Dynesty searches and completely eliminates the nuclear mode functionality that is no longer used.

- Removed checkpoint file deletion if statements from Nautilus and Dynesty search modules
- Eliminated nuclear mode logic from abstract search, paths, and grid search modules


Jammy2211:
The final push puts the if statements deleting checkpoint files back in, but now does so in the `output_search_internal` function.

I remembered that these if loops were used to delete checkpoint files that were a duplicate of the `search_internal.dill` file that normalizes the search. Putting them where they now are fixes the bug but retains the expected behaviour.

PR #1127, opened 2025-04-02
feature/csv error columns

### LINKED ISSUE #1087: Aggregator for csv building

rhayes777:
Resolves #1087

New usage:

```python

aggregate_csv.add_variable(
    "galaxies.lens.bulge.centre.centre_0",
    value_types=[ValueType.Median, ValueType.ValueAt1Sigma],
)
```

Median is included by default


Jammy2211:
I ran the example here and most functionality works:

https://github.com/Jammy2211/autogalaxy_workspace/blob/main/scripts/results/workflow/csv_make.py

However, the `values_at_sigma_1` (and I think the `values_at_sigma_3` are incorrect in `samples_summary.json`.

For example, I make the following .csv file:

```
"""
__Errors__

We can also output PDF values at a given sigma confidence of each parameter to the .csv file, using 
the `use_values_at_sigma` input and specifying the sigma confidence.

Below, we add the values at 3.0 sigma confidence to the .csv file, in order to compute the errors you would 
subtract the median value from these values.

The method below adds two columns to the .csv file, corresponding to the values at the lower and upper sigma values.
"""
agg_csv = af.AggregateCSV(aggregator=agg)

agg_csv.add_variable(
    argument="galaxies.galaxy.bulge.effective_radius",
    name="bulge_effective_radius_sigma",
    value_types=[af.ValueType.Median, af.ValueType.ValuesAt3Sigma]
)

agg_csv.save(path=workflow_path / "csv_simple_errors.csv")
```

The .csv file is:

```
id,bulge_effective_radius_sigma,bulge_effective_radius_sigma_lower_1_sigma,bulge_effective_radius_sigma_upper_1_sigma
441ad79fb6081c6f9ee07d3fff9eb838,1.4477396552296486,0.0003599749979079485,0.0015501703932636415
ab8a6e5d61cf0a5644e34235161bb658,1.4070892640122448,-0.002349170789082214,-0.0010475369447976472
```

The values for `bulge_effective_radius_sigma_lower_1_sigma,bulge_effective_radius_sigma_upper_1_sigma` are tiny, 0.001, when they should be much larger, heres one example of them from `model.results`:

```
            effective_radius                                                    1.6688 (1.6329, 1.7025)
```

The values in the .csv file are in the `samples_summary.json`, so the problem is how the values are being output theres as opposed to `AggregateCSV` itself.

Jammy2211:
I have confirmed that the correct values are in `samples_summary.json`, but that it is simply associating the wrong parameter to that value.

This implies that the _order_ of the values in the `samples_summary.json` file (which are stored in a list) is not consistent with the model in some way.

rhayes777:
> I have confirmed that the correct values are in `samples_summary.json`, but that it is simply associating the wrong parameter to that value.
> 
> This implies that the _order_ of the values in the `samples_summary.json` file (which are stored in a list) is not consistent with the model in some way.

Ah that's not good. This is why we need a dict really

rhayes777:
> Is there a reason why getting the `dict` thats in `samples_summary.json` in this format for the errors is diffcult, given we already have it for the median_pdf?
> 
> ```
>     "median_pdf_sample": {
>             "type": "instance",
>             "class_path": "autofit.non_linear.samples.sample.Sample",
>             "arguments": {
>                 "log_likelihood": -3761.9537991206726,
>                 "log_prior": 0.3401980233075243,
>                 "weight": 1.0,
>                 "kwargs": {
>                     "type": "dict",
>                     "arguments": {
>                         "galaxies.galaxy.bulge.ell_comps.ell_comps_0": 0.10795954960076128,
>                         "galaxies.galaxy.bulge.ell_comps.ell_comps_1": 0.07779982311130317,
>                         "galaxies.galaxy.bulge.effective_radius": 2.38834396240422,
>                         "galaxies.galaxy.bulge.sersic_index": 3.59010139964543,
>                         "galaxies.galaxy.disk.centre.centre_0": 0.0454336546817012,
>                         "galaxies.galaxy.disk.centre.centre_1": 0.1027471736412032,
>                         "galaxies.galaxy.disk.ell_comps.ell_comps_0": 0.17483925334984612,
>                         "galaxies.galaxy.disk.ell_comps.ell_comps_1": 0.018374008497923464,
>                         "galaxies.galaxy.disk.effective_radius": 11.85494112289763
>                     }
>                 }
>             }
>         },
> ```

It means changing the interface of the Samples classes

rhayes777:
It's weird it definitely looks like things should always be ordered by prior id

Jammy2211 (approved):

PR #1129, opened 2025-04-13
fix unit tests by restoring .fits files and using MODEL_IMAGe

Jammy2211:
The following PR add a .csv visualization file for irregular source reconstructions:

https://github.com/Jammy2211/PyAutoLens/compare/feature/fits_reconstruction?expand=1

This PR adds a method to `AggregateFITS` which loads these .csv files as lists of dictionaries.

Its a bit odd it being in `AggregateFITS`, but doesn't fit with the usage of other `Aggregate` objects and I dont want to make a new one. There will be more .csv visuals in the future so I woill think about how to approach this going forward.


autofit/aggregator/summary/aggregate_fits.py
```diff
@@ -27,7 +28,7 @@ class FITSFit(Enum):
     The HDUs that can be extracted from the fit.fits file.
     """
 
-    ModelImage = "MODEL_IMAGE"
+    ModelData = "MODEL_IMAGE"
```
rhayes777, line 31: It's less confusing if name and value contain the same words but not absolutely necessary

rhayes777 (approved):

PR #1130, opened 2025-04-25
feature/graphical analysis output

### LINKED ISSUE #1114: Graphical Model 4: Hard Disk Output
The following example compares a run using analysis summing and graph:

**Do not run this using**  PYAUTOFIT_TEST_MODE=1 as you care about imaging output.

For analysis summing, we did a lot of work making the output to hard disk structured and easy to use:

Image

This has two main components:

- Each analysis has a subfolder `analysis_0`, which contains its specific images and `files` folder.
- There is a single unified `image` folder, containing images shared over all analysis objects like `subplot_dataset_combined.png`.

Graphical models do not have this bespoke output structure:

Image

Basically, can we move the visualuazation / file output specific tasks in analysis summing, which worked on each idnividual `Analysis` in the `analysis_list` and have it do this on graphical models? 

In hindsight, this is something we would always have wanted a graphical model to do :).

**Functions to move**

The code that needs moving is in `combined.py`: https://github.com/rhayes777/PyAutoFit/blob/main/autofit/non_linear/analysis/combined.py

We are just looking for elegent ways to get these `for_each_analysis` tools moved to graphical models.

rhayes777:
Resolves #1114 (pending full test)


rhayes777:
This appears to work having run the integration test

Jammy2211 (approved):

PR #1138, opened 2025-05-09
Feature/pyproject toml

Jammy2211:
Update to `project.toml` for support with most recent PEP updates on newer Python versions.


rhayes777 (approved):
Nice. I think you can also define stuff like what is a project import (i.e. other auto* libraries) so that automatic linters like ruff update your import order

PR #1150, opened 2025-06-23
Feature/truncated gaussian

Jammy2211:
### PR Description: Implementation of a Truncated Gaussian Prior with Hard Limits

Previously, limits on Gaussian priors (and other priors) were implemented as soft bounds defined by `lower_limit` and `upper_limit` attributes. When sampling from these priors, it was possible for drawn parameter values to fall outside the specified limits. In such cases, `autofit` performed a runtime check, and if the sampled value violated the limits, a `FitException` was raised, triggering a resampling of the model. This approach introduced inefficiencies due to potentially many rejected samples and less robust sampling behavior near the bounds.

---

### Changes Introduced by This PR

* **New `TruncatedGaussianPrior` class:**
  This class implements a Gaussian prior truncated between strict lower and upper bounds. Instead of sampling freely and then rejecting values outside the limits, this prior explicitly restricts the parameter space, normalizing the distribution within the truncated interval. This results in:

  * **Improved sampling efficiency:** Samples are guaranteed to lie within the specified bounds without the need for rejection.
  * **Better numerical stability and accuracy:** The probability density is properly normalized over the truncated interval, ensuring consistent behavior.

* **Support for prior passing:**
  The truncated Gaussian prior is now used in scenarios where prior parameters are updated or passed between model fits. For example, when a model-fit returns posterior estimates, these can be used to create new priors centered on the estimated median with limits reflecting credible intervals. Using a truncated Gaussian ensures the new prior accurately reflects the constrained parameter space without introducing invalid samples.

---

### Benefits and Impact

* Sampling within strict parameter bounds becomes more robust and efficient.
* Reduced risk of `FitException` errors due to out-of-bound samples.
* Enables more reliable chaining of priors during sequential inference or hierarchical modeling.
* Provides a cleaner, more interpretable API for users requiring truncated Gaussian priors.

### JAX

The primary motivation of this PR is JAX, as the resampling API required annoying if loops which are no longer present as they are built into parameter mapping.


test_autofit/aggregator/test_reference.py
```diff
@@ -84,11 +85,12 @@ def test_database_info(
     database_aggregator,
     output_directory,
 ):
+    print((output_directory / "database.info").read_text())
```
rhayes777, line 88: forgotten print statement

test_autofit/graphical/functionality/test_jacobians.py
```diff
@@ -1,152 +1,141 @@
...
+# def test_jacobian_equiv():
+#
+#     def linear(x, a, b, c):
+#         z = x.dot(a) + b
+#         return (z**2).sum(), z
+#
+#     x_, a_, b_, c_, z_ = variables("x, a, b, c, z")
+#     x = np.arange(10.0).reshape(5, 2)
+#     a = np.arange(2.0).reshape(2, 1)
+#     b = np.ones(1)
```
rhayes777, line 22: lol core mp tests out of action

rhayes777 (approved):
Nice how well does the truncated version work?

PR #1220, opened 2026-04-16
feat: add autofit.jax opt-in pytree registration API

Jammy2211:
## Summary
Introduce `autofit.jax.enable_pytrees()` and `autofit.jax.register_model(model)` so callers can register PyAutoFit prior/model classes — and the user classes referenced by `Model.cls` (e.g. `Gaussian`, `Galaxy`) — as JAX pytree nodes on demand.

PyAutoFit deliberately does **not** register these classes at import time (the Nov 2025 cleanup removed eager `register_pytree_node_class` calls to keep JAX an optional dependency). This PR re-enables the functionality without bringing the eager JAX import back: the registration is lazy and opt-in.

The killer feature is `register_model(model)` → `_build_instance_pytree_funcs(model)`, which partitions each `Model.cls` constructor's arguments:
- Constants from the model definition (e.g. `af.Constant`, `Galaxy(redshift=0.5)`) → JAX `aux_data`, so they stay as concrete Python values inside a `jax.jit` trace.
- Prior-derived arguments → `children`, so they become JAX tracers.

Without this partition, constants would become tracers and code like `sorted(galaxies, key=lambda g: g.redshift)` in `Tracer` would raise `TracerBoolConversionError` under JIT.

## API Changes
One new public sub-package `autofit.jax` with two functions: `enable_pytrees()` registers the library's prior/model classes with `jax.tree_util`; `register_model(model)` walks a `Model` / `Collection` and registers every `Model.cls` it finds.

Both are no-ops if JAX is missing. Both are idempotent — re-registering a class is silently tolerated. See full details below.

## Test Plan
- [x] `pytest test_autofit/` passes (1223 tests)
- [x] New `test_autofit/jax/test_enable_pytrees.py` covers: round-trip flatten/unflatten, JIT compatibility, constants-stay-static under JIT, idempotency, and `Collection` round-trip
- [x] Verified end-to-end in `autolens_workspace_developer/jax_profiling/imaging/mge.py` — `jax.jit(jax.vmap(full_pipeline_from_params))` runs with pytree inputs

<details>

### Added
- `autofit.jax` — new sub-package.
- `autofit.jax.enable_pytrees() -> bool` — registers `GaussianPrior`, `UniformPrior`, `LogGaussianPrior`, `LogUniformPrior`, `TruncatedGaussianPrior`, `Model`, `Collection`, `ModelInstance` with `jax.tree_util`. Returns `True` on success, `False` if JAX is missing. Idempotent.
- `autofit.jax.register_model(model) -> bool` — calls `enable_pytrees()` then walks `model` (a `Model` or `Collection`) and registers each concrete `Model.cls`. The per-class flatten partitions constants (from `model.direct_instance_tuples`) into `aux_data` and priors into `children`. Returns `True` on success, `False` if JAX is missing. Idempotent.

### Removed / Renamed / Changed Signature / Changed Behaviour
None.

### Migration
Previously, callers who wanted JAX integration had to hand-register classes themselves (e.g. `jax.tree_util.register_pytree_node_class(af.GaussianPrior)`). Now:

- Before: ad-hoc `jax.tree_util.register_pytree_node_class(...)` calls per class
- After: `import autofit as af; af.jax.register_model(model)` before crossing a `jit` / `vmap` boundary

Legacy hand-registration still works; `register_model` tolerates classes already registered.

</details>

### DIFF
--- a/autofit/jax/__init__.py
+++ b/autofit/jax/__init__.py
@@ -1,0 +1,16 @@
+"""Opt-in JAX integration helpers for PyAutoFit.
+
+The library defines pytree ``tree_flatten`` / ``tree_unflatten`` methods on
+its model and prior classes but, by design, does NOT register them with JAX
+at import time. Eager registration would force ``jax.tree_util`` to load on
+every ``import autofit`` and reintroduce the heavy JAX import that the
+2025-11 cleanup removed.
+
+Call :func:`enable_pytrees` once before crossing a ``jax.jit`` or
+``jax.vmap`` boundary with PyAutoFit objects. The function is a no-op if JAX
+is not installed.
+"""
+
+from .pytrees import enable_pytrees, register_model
+
+__all__ = ["enable_pytrees", "register_model"]
--- a/autofit/jax/pytrees.py
+++ b/autofit/jax/pytrees.py
@@ -1,0 +1,150 @@
+"""Lazy JAX pytree registration for PyAutoFit's model and prior classes.
+
+The classes themselves already carry the necessary ``tree_flatten`` /
+``tree_unflatten`` methods. This module simply registers them with
+``jax.tree_util`` on demand, via the lazy autoconf wrapper, so callers can
+pass ``Model`` / ``Collection`` / ``ModelInstance`` / prior instances
+through ``jax.jit`` and ``jax.vmap`` directly.
+"""
+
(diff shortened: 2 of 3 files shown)


Jammy2211:
Workspace PR: https://github.com/PyAutoLabs/autolens_workspace_developer/pull/11

PR #1221, opened 2026-04-17
fix(jax): route concrete kwarg attrs to aux_data in pytree flatten

Jammy2211:
## Summary

Fixes two bugs in `autofit.jax._build_instance_pytree_funcs` that surfaced when migrating `jax_profiling/imaging/pixelization.py` to pass a pytree `ModelInstance` into JIT/vmap directly (follow-up to PyAutoLabs/autolens_workspace_developer#10, the same work that produced PyAutoFit#1220 for MGE):

1. **Concrete non-Prior kwargs became JAX tracers.** A `Galaxy(pixelization=<concrete Pixelization>)` attribute is stored directly on `model.__dict__` but is not typed as `Prior`/`AbstractPriorModel`. The old classification path (`direct_argument_names` / `direct_instance_tuples`) never saw it, so it was routed into `children` instead of `aux_data`. Under `jax.jit` this turned the pixelization into a tracer, causing `isinstance(x, aa.Pixelization)` inside `FitImaging` to silently return `False` and `fit.inversion` to resolve to `None`.
2. **`**kwargs` classes shared a broken flatten.** `Galaxy.__init__(self, redshift, **kwargs)` produces instances with different attribute sets per-model (lens: `bulge/mass/shear`; source: `pixelization`). JAX registers one flatten per class, so the first model's captured name list blew up with `AttributeError: 'Galaxy' object has no attribute 'bulge'` when flattening the source instance.

## API Changes

None — internal changes only. `enable_pytrees()` and `register_model(model)` signatures and semantics are unchanged; these are pure correctness fixes to the internal flatten/unflatten closures registered with `jax.tree_util`.

## Test Plan

- [x] New regression test `test_register_model_keeps_kwarg_constants_static` — exercises the exact `Galaxy(pixelization=...)` pattern: a concrete kwarg must stay identity-equal to the original object under JIT and pass `isinstance`.
- [x] Existing `test_register_model_keeps_constants_static` (Galaxy-redshift-style constant) still passes.
- [x] Full PyAutoFit suite green: `pytest test_autofit/ -x` → 1224 passed.
- [x] End-to-end smoke: `jax_profiling/imaging/pixelization.py` runs cleanly, both correctness assertions pass (step-by-step `log_evidence` matches reference within `rtol=1e-4`; vmap batch of 3 matches single-JIT result).

<details>

### Removed
None.

### Added
None.

### Renamed
None.

### Changed Signature
None.

### Changed Behaviour
- Internal only: `autofit.jax._build_instance_pytree_funcs` is now instance-driven. `flatten` iterates `vars(instance)` at call time against a shared `_CLASS_FIELD_CLASSIFIERS[cls]` dict populated across all `register_model` walks (rather than closing over a fixed name list captured from a single model at registration time). Attributes unknown to the classifier default to `aux_data` (safer than tracing an unknown object).

### Migration
No migration required. Callers of `autofit.jax.enable_pytrees` / `autofit.jax.register_model` see only a behavioural fix — patterns that previously silently produced broken JAX traces now work correctly.

</details>

### DIFF
--- a/autofit/jax/pytrees.py
+++ b/autofit/jax/pytrees.py
@@ -11,6 +11,8 @@ from autoconf.jax_wrapper import register_pytree_node, register_pytree_node_clas
 
 _ENABLED = False
 _REGISTERED_INSTANCE_CLASSES: set = set()
+_CLASS_FIELD_CLASSIFIERS: dict = {}
+_CLASS_CONSTRUCTOR_ARGS: dict = {}
 
 
 def enable_pytrees() -> bool:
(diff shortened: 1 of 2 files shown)


Jammy2211:
Workspace PR: https://github.com/PyAutoLabs/autolens_workspace_developer/pull/12

PR #1229, opened 2026-04-19
feat: lazily JIT-cache fit_for_visualization when flag set

Jammy2211:
## Summary
Replace the pass-through stub in `Analysis.fit_for_visualization` with a lazy `jax.jit` cache. With `use_jax_for_visualization=True`, the first call compiles `self.fit_from` and stores it on the instance as `_jitted_fit_from`; subsequent calls reuse the compiled function. Default behaviour (flag off) is unchanged — this is a controlled opt-in ahead of the remaining `Fit*` pytree registrations.

## API Changes
`Analysis.fit_for_visualization` now lazily constructs and caches `jax.jit(self.fit_from)` on the instance when `use_jax_for_visualization=True`. Flag-off behaviour and all public signatures are unchanged. See full details below.

## Test Plan
- [x] `test_autofit/analysis/test_use_jax_for_visualization.py` updated to exercise both the JIT path (asserts `_jitted_fit_from` is created on first call and reused on second) and the no-flag path
- [x] Full `test_autofit/` suite passes (1232 tests)
- [x] End-to-end validated via `modeling_visualization_jit.py` in the linked workspace PR: 7.6× speedup compile → cached, Nautilus live-search fires the JIT path and writes `subplot_fit.png`

<details>

### Changed Behaviour
- `Analysis.fit_for_visualization(instance)` — when `use_jax_for_visualization=True` on the instance, lazily constructs `jax.jit(self.fit_from)` on the first call, caches it as `self._jitted_fit_from`, and returns the compiled function's output for every subsequent call. Falls back to plain `self.fit_from(instance)` when the flag is off.

### Added (internal)
- `Analysis._jitted_fit_from` — lazy-initialized attribute; `None` until the first `fit_for_visualization` call with the flag on.

### Migration
No migration needed for flag-off callers (the default). Callers that set `use_jax_for_visualization=True` now get a JIT-compiled `fit_from` — their `Fit*` return type and all nested types must be pytree-registered or the compile will error. See `admin_jammy/prompt/autolens/fit_*_pytree_*.md` for the variant matrix.

</details>

### DIFF
--- a/autofit/non_linear/analysis/analysis.py
+++ b/autofit/non_linear/analysis/analysis.py
@@ -58,22 +58,43 @@ class Analysis(ABC):
         """
         Build the fit used by the visualizer.
 
-        Currently a thin dispatch over ``self.fit_from``: when ``use_jax=True``
-        the fit is built on the eager JAX path (``self._xp is jnp``) and the
-        plotter materialises arrays to NumPy at the matplotlib boundary. The
-        ``use_jax_for_visualization`` flag is an explicit opt-in today — it
-        marks the intent to use JAX for visualization, and is the dispatch
-        point where full ``jax.jit``-wrapping will plug in once
-        :class:`autolens.imaging.fit_imaging.FitImaging` and its nested
-        autoarray types are registered as JAX pytrees (that work is tracked
-        separately — see ``admin_jammy/prompt/autolens/fit_imaging_pytree.md``).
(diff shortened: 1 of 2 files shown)


Jammy2211:
Workspace PR: https://github.com/PyAutoLabs/autolens_workspace_test/pull/33

PR #1292, opened 2026-05-22
fix: PYAUTO_TEST_MODE should write to a separate output dir

### LINKED ISSUE #1291: fix: PYAUTO_TEST_MODE should write to a separate output dir
## Overview

When a script is smoke-tested under `PYAUTO_TEST_MODE` and then later run for real, AutoFit silently returns the cached test-mode result ("Fit Already Completed: skipping non-linear search") because the output path is identical between the two runs. The proper run never happens — symptom is every recovered σ on the supposedly-real run being exactly 0.

This was hit on the IC50 EP / graphical scripts. Current workaround is `rm -rf output/<path-prefix>/` before the first proper run, which is runtime discipline contributors will forget. The structural fix is to namespace the test-mode output path so smoke and prod outputs are physically co-resident.

## Plan

- Add a small helper in `autofit/non_linear/paths/abstract.py` that returns `"test_mode"` when `PYAUTO_TEST_MODE` is set in the environment, else `None`.
- Filter that segment into the path composition at every site that builds an output path from `conf.instance.output_path` (three sites — see detailed plan).
- Net effect: prod writes to `output/<prefix>/<name>/<id>/`, test writes to `output/test_mode/<prefix>/<name>/<id>/`. Caches can no longer collide.
- Add a PyAutoFit regression test that runs a search under `PYAUTO_TEST_MODE=2`, then again with it unset, and asserts the second run re-fits.
- Validation sweep: `/health_check` + per-workspace `/smoke_test` across every workspace, including `ic50_workspace` where the bug was first hit.

<details>
<summary>Detailed implementation plan</summary>

### Affected Repositories
- PyAutoFit (primary, code change)
- All workspaces (validation sweep only — no workspace edits): autofit_workspace, autogalaxy_workspace, autolens_workspace, autolens_workspace_test, autogalaxy_workspace_test, euclid_strong_lens_modeling_pipeline, ic50_workspace

### Work Classification
Library (PyAutoFit) — single-repo code change with cross-workspace validation. No workspace code edits.

### Branch Survey

| Repository | Current Branch | Dirty? |
|-----------|---------------|--------|
| ./PyAutoFit | main | clean |
| ./PyAutoPrompt | main | clean (new prompt file untracked, expected) |

**Suggested branch:** `feature/test-mode-output-path`
**Worktree root:** `~/Code/PyAutoLabs-wt/test-mode-output-path/` (created by `/start_library`)

No conflicts: `worktree_check_conflict test-mode-output-path PyAutoFit` exits 0. The `smoke-test-optimization` entry in active.md touches PyAutoFit test_mode skip-guards but not the paths module, so the surface area doesn't overlap.

### Implementation Steps

1. **Add the helper** in `autofit/non_linear/paths/abstract.py`, near the top of the module:
   ```python
   def _test_mode_segment() -> Optional[str]:
       """Returns 'test_mode' when PYAUTO_TEST_MODE is set, else None."""
       return "test_mode" if os.environ.get("PYAUTO_TEST_MODE") else None
   ```
2. **Patch the three composition sites** to filter the segment in right after `conf.instance.output_path`:
   - `autofit/non_linear/paths/abstract.py:239` — `output_path` property on `AbstractPaths`.
   - `autofit/non_linear/paths/directory.py:533` — `_make_path()` on `DirectoryPaths`. (Note: this site already omits `unique_tag` that the property includes — pre-existing inconsistency, leave untouched.)
   - `autofit/database/__init__.py:60` — read context first; patch only if it represents a search-output dir.
3. **Confirm `SubDirectoryPaths` is covered for free** — its `_output_path` delegates to `self.parent.output_path`, so the parent fix propagates. Verify by reading `sub_directory_paths.py:31-79,123-128`.
4. **Add regression test** in `test_autofit/non_linear/paths/test_paths.py`:
   - Construct `DirectoryPaths(name=..., path_prefix=...)` with `PYAUTO_TEST_MODE=2` set in the env, assert the resolved `output_path` contains `test_mode` as a path component.
   - Repeat with the env var unset, assert it does not.
   - Use `monkeypatch.setenv` / `monkeypatch.delenv` so the test is isolated.
5. **Document** the new path layout in the relevant docstring (the `AbstractPaths.__init__` docstring already includes a worked example — add a one-paragraph note that `PYAUTO_TEST_MODE` prepends `test_mode/`).
6. **Run PyAutoFit unit suite** — must be green.
7. **Cross-workspace validation sweep**:
   - `/health_check` (fast-forward all repos, run library unit tests + workspace smoke).
   - Targeted `/smoke_test` per workspace: autofit_workspace, autogalaxy_workspace, autolens_workspace, autolens_workspace_test, autogalaxy_workspace_test, euclid_strong_lens_modeling_pipeline, ic50_workspace.
8. **Manual cross-check** on one or two scripts that already have output in `output/`: run with `PYAUTO_TEST_MODE=2`, confirm the artefact lands under `output/test_mode/...`, then run unset and confirm the second invocation actually runs the sampler (no "Fit Already Completed").

### Key Files
- `PyAutoFit/autofit/non_linear/paths/abstract.py` — helper + property patch.
- `PyAutoFit/autofit/non_linear/paths/directory.py` — `_make_path()` patch.
- `PyAutoFit/autofit/database/__init__.py` — possibly patch, depends on context.
- `PyAutoFit/test_autofit/non_linear/paths/test_paths.py` — regression test.

### Post-merge follow-up
Update memory `feedback_autofit_cache_resume_pyauto_test_mode`: the `rm -rf` workaround is obsolete once this ships.

</details>

## Original Prompt

<details>
<summary>Click to expand starting prompt</summary>

# PYAUTO_TEST_MODE should mutate the AutoFit output path

When a script is smoke-tested under `PYAUTO_TEST_MODE` and then later run for real, AutoFit silently returns the cached test-mode result ("Fit Already Completed: skipping non-linear search") because the output path is identical between the two runs. The proper run never happens — symptom is every recovered σ on the supposedly-real run being exactly 0.

This was hit on the IC50 EP / graphical scripts and the current workaround is "rm -rf `output/<path-prefix>/` before the first proper run", which is runtime discipline that contributors will forget.

## What I want

Make `PYAUTO_TEST_MODE` mutate the output path so smoke and prod outputs are physically co-resident and can't collide. The `rm -rf` workaround becomes unnecessary.

## Where

Inject a `test_mode/` segment at the root of the output tree when `os.environ.get("PYAUTO_TEST_MODE")` is set and truthy:

- prod: `output/<prefix>/<name>/<identifier>/`
- test: `output/test_mode/<prefix>/<name>/<identifier>/`

I prefer collecting all test-mode outputs under one tree (easy to gitignore or wipe wholesale) rather than splitting at the identifier level.

There are **three** places that compose paths from `conf.instance.output_path` — they all need the same treatment, so factor a shared helper rather than patching each site:

1. `PyAutoFit/autofit/non_linear/paths/abstract.py:239` — `output_path` property on `AbstractPaths`.
2. `PyAutoFit/autofit/non_linear/paths/directory.py:533` — `_make_path()` on `DirectoryPaths`, which reconstructs the path manually instead of using the property. (Aside: it omits `unique_tag` while the property includes it — pre-existing inconsistency, don't fix here.)
3. `PyAutoFit/autofit/database/__init__.py:60` — read the surrounding context before patching; may or may not need the segment depending on whether it represents a search-output dir.

`SubDirectoryPaths` overrides `output_path` in `sub_directory_paths.py:74,123` but its `_output_path` delegates to `self.parent.output_path`, so it inherits the fix for free via the parent.

Suggested helper signature:

```python
def _test_mode_segment() -> Optional[str]:
    return "test_mode" if os.environ.get("PYAUTO_TEST_MODE") else None
```

Then filter it into the path components at each of the three sites, right after `conf.instance.output_path`.

## Validation

The code change is small; the cost is validation across every workspace, since the risk is a script that secretly relies on smoke and prod sharing the output path.

- PyAutoFit unit suite green.
- Add a regression test: run a search under `PYAUTO_TEST_MODE=2`, then again with `PYAUTO_TEST_MODE` unset, and assert the second run re-fits (does not hit the `is_complete` short-circuit).
- `/health_check` followed by `/smoke_test` on every workspace: `autofit_workspace`, `autogalaxy_workspace`, `autolens_workspace`, `autolens_workspace_test`, `autogalaxy_workspace_test`, `euclid_strong_lens_modeling_pipeline`, `ic50_workspace`. ic50 is where the bug was first hit so it must be in the sweep.
- Pick one or two scripts that already have output in `output/`, run them with and without `PYAUTO_TEST_MODE`, and confirm the two land in distinct directories.

## Out of scope

- Renaming smoke runs script-by-script (the whole point is the per-script rename is unnecessary).
- Touching the `is_complete` short-circuit — once paths differ, the existing logic is correct.

## Once shipped

Update the `feedback_autofit_cache_resume_pyauto_test_mode` memory: the `rm -rf` workaround is obsolete.

</details>

Jammy2211:
## Summary
When a script is smoke-tested under `PYAUTO_TEST_MODE` and then run for real, AutoFit silently returns the cached test-mode result ("Fit Already Completed: skipping non-linear search") because the output path is identical between the two runs. The proper run never happens — symptom is every recovered σ on the supposedly-real run being exactly 0.

This PR namespaces the test-mode output path so smoke and prod outputs are physically co-resident under `output/test_mode/...` and cannot collide. The previous workaround — `rm -rf output/<path-prefix>/` before the first proper run — is no longer required.

## API Changes
No public symbol additions, removals, or signature changes.

Behaviour change: when the `PYAUTO_TEST_MODE` environment variable is set (any truthy value), AutoFit search output paths get a `test_mode/` segment inserted directly after `conf.instance.output_path`. This affects every search-output path including SQLite database files opened via `open_database()` with a relative filename.

See full details below.

## Test Plan
- [x] New regression tests in `test_autofit/non_linear/paths/test_paths.py` cover both `output_path` and `_make_path` with `PYAUTO_TEST_MODE` set vs unset (4 tests, all pass)
- [x] Full PyAutoFit unit suite: 1408 passed / 1 skipped / 0 failed
- [ ] Validation sweep across workspaces (`/health_check` + per-workspace `/smoke_test` on autofit/autogalaxy/autolens workspaces + ic50_workspace) — see linked issue. Risk being checked: any script that secretly relied on smoke and prod sharing an output path

Closes #1291

<details>

### Removed
None.

### Added
- `autofit.non_linear.paths.abstract._test_mode_segment()` — private module-level helper returning `"test_mode"` when `PYAUTO_TEST_MODE` env var is set, else `None`.

### Renamed
None.

### Changed Signature
None.

### Changed Behaviour
- `AbstractPaths.output_path` property — when `PYAUTO_TEST_MODE` is set, the returned path includes a `test_mode/` segment right after the configured output root.
- `DirectoryPaths._make_path()` — same `test_mode/` segment injection when env var is set.
- `autofit.database.open_database(filename)` — when called with a relative `filename` ending in `.sqlite`, the resolved path is prefixed with `test_mode/` under the output root when env var is set. Absolute filenames are unchanged.

### Migration
No code changes required for callers. After this lands, you can stop running `rm -rf output/<path-prefix>/` between a smoke run and a real run at the same paths.

</details>

### DIFF
--- a/autofit/database/__init__.py
+++ b/autofit/database/__init__.py
@@ -7,6 +7,7 @@ from autoconf import conf
 from .aggregator import *
 from .migration.steps import migrator
 from .model import *
+from autofit.non_linear.paths.abstract import _test_mode_segment
 
 
 def open_database(
@@ -60,6 +61,9 @@ def open_database(
         output_path = conf.instance.output_path
 
         if not filename.startswith("/"):
+            segment = _test_mode_segment()
+            if segment:
+                output_path = f"{output_path}/{segment}"
             filename = f"{output_path}/{filename}"
 
         os.makedirs(
--- a/autofit/non_linear/paths/abstract.py
+++ b/autofit/non_linear/paths/abstract.py
@@ -22,6 +22,19 @@ logger = logging.getLogger(__name__)
 pattern = re.compile(r"(?<!^)(?=[A-Z])")
 
 
+def _test_mode_segment() -> Optional[str]:
+    """
+    Returns ``"test_mode"`` when ``PYAUTO_TEST_MODE`` is set in the
+    environment, else ``None``.
+
+    Inserted into the output-path composition so that smoke runs land
+    under ``output/test_mode/...`` instead of sharing a directory with
+    real runs. Without this, a cached test-mode result short-circuits
+    a later real run at the same paths ("Fit Already Completed").
+    """
+    return "test_mode" if os.environ.get("PYAUTO_TEST_MODE") else None
+
+
 class AbstractPaths(ABC):
     def __init__(
         self,
@@ -52,6 +65,15 @@ class AbstractPaths(ABC):
         The output path of the `NonLinearSearch` results will be:
 
         /path/to/output/folder_0/folder_1/name
+
+        If the ``PYAUTO_TEST_MODE`` environment variable is set, a
+        ``test_mode`` segment is inserted directly after the output
+        root:
+
+        /path/to/output/test_mode/folder_0/folder_1/name
+
+        This keeps smoke-test artefacts in a sibling tree so they
+        cannot be picked up by a later real run with the same paths.
 
         Parameters
         ----------
@@ -246,6 +268,7 @@ class AbstractPaths(ABC):
                 None,
                 [
                     str(conf.instance.output_path),
+                    _test_mode_segment(),
                     str(self.path_prefix),
                     self.unique_tag,
                     str(self.name),
--- a/autofit/non_linear/paths/directory.py
+++ b/autofit/non_linear/paths/directory.py
@@ -16,7 +16,7 @@ from autofit.text import formatter
 from autofit.tools.util import open_
 from autofit.non_linear.samples.samples import Samples
 
-from .abstract import AbstractPaths
+from .abstract import AbstractPaths, _test_mode_segment
 
 from ..samples import load_from_table
 from autofit.non_linear.samples.pdf import SamplesPDF
@@ -530,7 +530,11 @@ class DirectoryPaths(AbstractPaths):
         The path terminates with the identifier, unless the identifier has already
         been added to the path.
         """
-        path_ = Path(conf.instance.output_path) / self.path_prefix / self.name
+        path_ = Path(conf.instance.output_path)
+        segment = _test_mode_segment()
+        if segment:
+            path_ = path_ / segment
+        path_ = path_ / self.path_prefix / self.name
         if self.is_identifier_in_paths:
             path_ = path_ / self.identifier
         return path_
--- a/test_autofit/non_linear/paths/test_paths.py
+++ b/test_autofit/non_linear/paths/test_paths.py
@@ -75,3 +75,33 @@ def test_serialize(model):
 def test_unique_tag():
     paths = af.DirectoryPaths(unique_tag="unique_tag")
     assert "unique_tag" in paths.output_path.parts
+
+
+class TestTestModeOutputPath:
+    """
+    PYAUTO_TEST_MODE must namespace the AutoFit output path so smoke
+    runs cannot poison the cache for a later real run at the same
+    paths. Regression for the "Fit Already Completed" silent skip.
+    """
+
+    def test_output_path_contains_test_mode_segment_when_env_set(
+        self, monkeypatch
+    ):
+        monkeypatch.setenv("PYAUTO_TEST_MODE", "2")
+        paths = af.DirectoryPaths(name="name", path_prefix="prefix")
+        assert "test_mode" in paths.output_path.parts
+
+    def test_output_path_excludes_segment_when_env_unset(self, monkeypatch):
+        monkeypatch.delenv("PYAUTO_TEST_MODE", raising=False)
+        paths = af.DirectoryPaths(name="name", path_prefix="prefix")
+        assert "test_mode" not in paths.output_path.parts
+
+    def test_make_path_contains_segment_when_env_set(self, monkeypatch):
+        monkeypatch.setenv("PYAUTO_TEST_MODE", "2")
+        paths = af.DirectoryPaths(name="name", path_prefix="prefix")
+        assert "test_mode" in paths._make_path().parts
+
+    def test_make_path_excludes_segment_when_env_unset(self, monkeypatch):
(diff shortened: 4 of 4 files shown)


Jammy2211:
Workspace PRs:
- https://github.com/PyAutoLabs/autolens_workspace_test/pull/119
- https://github.com/PyAutoLabs/autogalaxy_workspace_test/pull/61

Issue #1307, opened 2026-06-01
feat: cross-Analysis shared per-evaluation state in FactorGraphModel (autofit + 1D toy)


Jammy2211:
> **Sub-task A** of the `analysis_shared_state` epic (tracker: `PyAutoPrompt/z_features/analysis_shared_state.md`).
> This issue covers the **autofit deliverable** — the generic mechanism, the 1D Gaussian toy, and its fast tests.
> The lensing datacube consumer is **sub-task B**, tracked separately and issued later
> (`PyAutoPrompt/autolens/datacube_shared_state_consumer.md`).

## Overview

`FactorGraphModel` fits N per-factor `Analysis` objects that share a model, summing
their likelihoods. When the factors share model parameters, a large fraction of each
factor's likelihood is *identical work* recomputed N times. This task gives PyAutoFit a
domain-agnostic mechanism for a factor to compute a per-evaluation shared object **once**
and have every factor reuse it.

The work is built **toy-first**: a 1D Gaussian shared-state example (the natural extension
of `af.ex` + `scripts/features/graphical_models.py`) is the development vehicle and proof,
paired with fast, deterministic tests in `autofit_workspace_test`. Proving the mechanism on
the toy here unblocks the motivating lensing consumer (ALMA datacube; sub-task B), where
~97% of the per-channel inversion-setup work is channel-invariant and currently rebuilt N
times. **PyAutoFit ships only the generic, domain-agnostic protocol** — no lensing logic.

## Plan (autofit deliverable)

**Phase 1 — PyAutoFit mechanism + toy `af.ex` pieces (library, Opus)**
- Add opt-in `Analysis.shared_state_from(instance) -> None` (default `None`, sibling of `modify_before_fit`)
  and a defaulted `shared=` kwarg on `log_likelihood_function`.
- `FactorGraphModel` computes the shared object once from the **lead factor** before the loop and forwards
  it to each factor **only when non-`None`** → existing graphs byte-for-byte unchanged.
- Add a **shared-aware example `Analysis`** to `autofit/example/analysis.py` so the toy is reusable as
  `af.ex` by both the workspace tutorial and the fast tests.
- Unit tests in `test_autofit/graphical/`.

**Phase 2 — autofit_workspace tutorial + docs (workspace, Opus prose)**
- New `scripts/features/shared_analysis_state.py` extending `graphical_models.py`, framing the shared
[...]

PR #1308, opened 2026-06-01
feat: cross-Analysis shared per-evaluation state in FactorGraphModel

Jammy2211:
## Summary

Adds an opt-in mechanism for the per-factor `Analysis` objects of a `FactorGraphModel` to compute a model-dependent object **once per likelihood evaluation** and reuse it across every factor, instead of each factor recomputing identical work. This is the generic, domain-agnostic foundation for the ALMA datacube likelihood (sub-task B, tracked separately), where ~97% of each channel's inversion-setup work is identical across channels and currently rebuilt N times.

The design mirrors the existing `EPAnalysisFactor` precedent (graph-level state attached around the factor loop) and `Analysis.modify_before_fit` (the precompute-then-reuse hook), but runs **once per likelihood evaluation** rather than once per fit. It is fully backward compatible: when no factor opts in, every graph behaves exactly as before.

Resolves part of #1307 (autofit deliverable; the 1D Gaussian workspace tutorial and fast tests follow).

## API Changes

All changes are additive and backward compatible — no removals or renames.

- Added `Analysis.shared_state_from(instance) -> None`: an opt-in hook (default returns `None`) that computes a per-evaluation object shared across a `FactorGraphModel`'s factors. The per-evaluation, cross-factor sibling of `modify_before_fit`.
- `Analysis.log_likelihood_function` (and the `AnalysisFactor` / `PriorFactor` / `HierarchicalFactor` variants) gain an optional `shared=None` keyword argument. Existing analyses that do not declare it keep working because `FactorGraphModel` forwards `shared` only when a factor opts in.
- `af.ex.Analysis` gains an opt-in `share_model_data=False` constructor flag plus a worked `shared_state_from` implementation, demonstrating the mechanism on the 1D Gaussian toy.

See full details below.

## Test Plan

- [ ] `python -m pytest test_autofit/graphical/` — 188 existing + 4 new tests pass
- [ ] New `test_autofit/graphical/test_shared_state.py`:
  - shared state computed once per evaluation (not once per factor)
  - shared-path likelihood exactly equals the unshared per-factor sum
  - a graph with no opt-in factor is unchanged
  - the default `shared_state_from` returns `None`

(Note: the full `test_autofit/` suite has 13 pre-existing failures under `test_autofit/non_linear/search/nest/nss/` that are unrelated to this change and already broken on `main`.)

<details>

### Added
- `autofit.non_linear.analysis.analysis.Analysis.shared_state_from(self, instance)` — returns `None` by default; override to share per-evaluation state across `FactorGraphModel` factors.
- `autofit.example.analysis.Analysis.__init__(..., share_model_data=False)` — opt-in flag enabling the shared-state path on the 1D Gaussian example.
- `autofit.example.analysis.Analysis.shared_state_from(self, instance)` — worked example: computes the shared model data once when `share_model_data=True`.

### Changed Signature (backward compatible — new optional kwarg with default)
- `Analysis.log_likelihood_function(self, instance)` → `(self, instance, shared=None)`
- `AnalysisFactor.log_likelihood_function(self, instance)` → `(self, instance, shared=None)`
- `PriorFactor.log_likelihood_function(self, instance)` → `(self, instance, shared=None)` (accepts and ignores `shared`)
- `HierarchicalFactor` family `log_likelihood_function(self, instance)` → `(self, instance, shared=None)` (accepts and ignores `shared`)
- `autofit.example.analysis.Analysis.log_likelihood_function(self, instance, xp=np)` → `(self, instance, shared=None, xp=np)`

### Changed Behaviour
- `FactorGraphModel.log_likelihood_function` now asks the lead factor for a shared object (`_shared_state_from`) before the per-factor loop and forwards it to each factor when non-`None`. When no factor opts in (the default), behaviour is unchanged.

### Migration
- None required. Existing `Analysis` subclasses and graphs work unchanged. To opt in, override `shared_state_from` to return a shared object and have `log_likelihood_function(self, instance, shared=None)` consume it (falling back to its own computation when `shared is None`).

</details>

### DIFF
--- a/autofit/example/analysis.py
+++ b/autofit/example/analysis.py
@@ -36,7 +36,13 @@ class Analysis(af.Analysis):
 
     LATENT_KEYS = ["gaussian.fwhm"]
 
-    def __init__(self, data: np.ndarray, noise_map: np.ndarray, use_jax=False):
+    def __init__(
+        self,
+        data: np.ndarray,
+        noise_map: np.ndarray,
+        use_jax=False,
+        share_model_data=False,
+    ):
         """
         In this example the `Analysis` object only contains the data and noise-map. It can be easily extended,
         for more complex data-sets and model fitting problems.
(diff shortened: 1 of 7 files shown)


Jammy2211:
Workspace PR: https://github.com/PyAutoLabs/autofit_workspace/pull/69

Jammy2211:
Workspace PR: https://github.com/PyAutoLabs/autofit_workspace_test/pull/31

PR #1408, opened 2026-07-21
fix: test-mode bypass tolerates a single-eval FitException (resample-equivalent)

Jammy2211:
## Summary

`TEST_MODE=2`'s `_fit_bypass_test_mode` (`autofit/non_linear/search/abstract_search.py`) calls the likelihood **once** to verify the analysis runs. If that single prior-median instance is pathological — a non-positive-definite inversion, or a degenerate Delaunay mesh that yields NaN vertices — the analysis raises `af.exc.FitException`. That is exactly the signal a real non-linear search absorbs by **resampling**. In test mode there is no sampler, so the exception hard-failed the whole run.

This made flaky pixelization scripts intermittently break CI smoke. Concretely, `autolens_workspace/scripts/interferometer/features/pixelization/delaunay.py` fails ~50% on CI py3.12 with `ValueError: Points cannot contain NaN` from the Delaunay/qhull triangulation, via `source_pix_2` → `_fit_bypass_test_mode` → `log_likelihood_function` → inversion → `scipy_delaunay` (autolens_workspace#307, PyAutoLens#640). The same workspace commit passed on a parallel run — a true nondeterministic (CI-thread/BLAS-dependent) flake.

## Change

Wrap the single verification eval: on `FitException`, log the cause and keep the existing `-1.0e99` sentinel (the same effect as a resample-to-reject). **Only `FitException` is caught** — genuine code errors (e.g. `ValueError`, `AttributeError`) still propagate and fail test mode, so a truly broken likelihood stays loud. Central fix: benefits every analysis (imaging, interferometer, galaxy), not just this one script. Complements PyAutoLens#639 (`raise_fit_exception` preserves the cause behind the FitException).

## Tests

- `TestBypassToleratesFitException::test__test_mode_2__fitexception_tolerated__fit_completes_with_sentinel` — fit completes, sentinel log-likelihood retained.
- `...::test__test_mode_2__non_fitexception_still_propagates` — a `ValueError` still raises.
- 108 `non_linear/{search,result,samples}` tests pass; no regressions.

## API Changes

None. Behaviour change is confined to the `TEST_MODE=2` bypass path; public API, signatures, and normal-search behaviour are unchanged. No downstream workspace migration required.

## Follow-ups

- autolens_workspace: un-park the interferometer Delaunay `no_run.yaml` entry + add to smoke (library-first: merge this PR first).
- Phase 2 (PyAutoArray): fix the actual NaN/non-PD producer (candidate: `fnnls.py:134` divide-by-zero; degenerate Hilbert-mesh vertices) so the fit is not merely tolerated but correct.

### DIFF
--- a/autofit/non_linear/search/abstract_search.py
+++ b/autofit/non_linear/search/abstract_search.py
@@ -872,9 +872,24 @@ class NonLinearSearch(AbstractFactorOptimiser, ABC):
         log_likelihood = -1.0e99
         if call_likelihood:
             instance = model.instance_from_vector(vector=parameter_vector)
-            log_likelihood = float(
-                analysis.log_likelihood_function(instance)
-            )
+            try:
+                log_likelihood = float(
+                    analysis.log_likelihood_function(instance)
+                )
+            except exc.FitException as e:
+                # A `FitException` means this particular instance is pathological
+                # (e.g. a non-positive-definite inversion, or a degenerate mesh
+                # that yields NaN vertices). In a real search the sampler absorbs
+                # this by resampling; test mode has no sampler, so a single
+                # unlucky verification eval must not hard-fail the run. Keep the
+                # `-1.0e99` sentinel — the same effect a resample-to-reject has —
+                # and log the cause so a genuinely broken likelihood stays visible.
+                # Only `FitException` is caught: real code errors still propagate.
+                logger.warning(
+                    "TEST MODE 2: likelihood verification raised FitException "
+                    f"({e.__cause__ or e!r}); treating as a resample-rejected "
+                    "instance and continuing with the sentinel log likelihood."
+                )
 
         sample_list = self._build_fake_samples(
             model=model,
--- a/test_autofit/non_linear/search/test_abstract_search.py
+++ b/test_autofit/non_linear/search/test_abstract_search.py
@@ -420,3 +420,56 @@ class TestBypassWritesCompleted:
         )
 
         assert result is not None
+
+
+class _FitExceptionAnalysis(af.m.MockAnalysis):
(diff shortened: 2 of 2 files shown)


Jammy2211:
**Design validation (parity with existing autofit conventions).** This isn't a bespoke workaround — `FitException → resample sentinel` is how autofit already handles pathological instances in three places, and this change makes the test-mode bypass consistent with them:

1. **Real numpy fitness wrapper** — `fitness.py:255-256`:
   ```python
   except exc.FitException:
       return self.resample_figure_of_merit
   ```
   The live sampler already treats a FitException as a resample-to-reject. The bypass is the only likelihood call that didn't.

2. **Sentinel value matches the searches these scripts use** — Nautilus and Dynesty set `resample_figure_of_merit=-1.0e99` (`nautilus/search.py:194,215`, `dynesty/.../abstract.py:184`). That is exactly the `-1.0e99` the bypass already defaults to (`abstract_search.py:872`) and that this fix retains — so a resample-rejected instance produces the identical sentinel in test mode and in a real Nautilus fit.

3. **Initializer** — `initializer.py:35,39` treats both `FitException` and `figure_of_merit < -1e98` as "invalid, resample" (`return None`), and its own test-mode bypass uses `-1.0e99` (`initializer.py:233`).

Scope is deliberately narrow (only `FitException`, matching `fitness.py`) so genuine code errors still fail test mode loudly. Complements PyAutoLens#639 (`raise_fit_exception` preserves the cause), whose body explicitly flagged the "single evaluation not absorbed by a sampler under PYAUTO_TEST_MODE" problem but scoped the tolerance out — this PR completes it.

CI: unittest 3.12 + 3.13 green, docs green.

